PR #14, opened 2019-04-28
Relase 2019-04-28

blckmn:


src/pages/news/betaflight-release-4.0-delayed.ejs
```diff
@@ -0,0 +1,53 @@
+---
+layout: article
+title: Betaflight 4.0 release delayed
```
mikeller: I think this is outdated and can be removed now that Betaflight 4.0 has been released.

blckmn: The news articles will kinda hang around but get superseded - think of it as an audit of announcements. It won't get front page as new articles are created.

PR #19, opened 2019-08-03
Fixed link in Betaflight 4.1 news item.

mikeller:


mikeller:
Has this been deployed? It's still not working. :-(

Also, can you find out if leaving the `link:` field empty will default it to the name of the ejs file?

blckmn:
Yeah I’ll make that change.

On 4 Aug 2019, at 08:54, Michael Keller <notifications@github.com> wrote:

Has this been deployed? It's still not working. :-(

Also, can you find out if leaving the link: field empty will default it to the name of the ejs file?

PR #32, opened 2019-08-18
Removing pilots, and members. 

blckmn:
Will look for alternative source for pilot details.


blckmn:
Moved the members to feed off Github, and the pilots needs a source of data - which will need to be sorted at some point.

McGiverGim:
@blckmn I see that you removed the pilots content, but I see the link at the top bar:

image

Is that expected?

mikeller:
@McGiverGim: I wonder if we are having regional synch issues again. I can still see the content for 'Pilots':

image

McGiverGim:
😆 

It seems... I don't see the Featured pilots content...

PR #50, opened 2019-10-26
Fixed OpenGraph data / logo for Facebook.

mikeller:


src/layouts/article.ejs
```diff
@@ -2,15 +2,17 @@
...
 		<meta name="description" content="">
 		<meta name="keywords" content="">
 		<meta charset="UTF-8">
-		<meta property="og:image" content="https://betaflight.com/img/small_logo.png">
+		<meta property="og:image" content="https://betaflight.com/img/logo_fb.png">
 		<meta property="og:description" content="<%= page.description %>">
 		<meta property="og:title" content="betaflight.com - <%= page.title %>">
-		
+		<meta property="og:url" content="https://betaflight.com/">
+		<meta property="og:type" content="website">
```
blckmn: The URL here should be the article url, however it’s not available at this point in nanogen. I’m submitting a PR there to fix that. So this’ll do in the interim.

mikeller: That is for `og:url`? Not entirely sure about that, according to https://developers.facebook.com/docs/sharing/webmasters/ it should be the 'canonical URL', and the 'likes and shares are aggreagated' bit (to me) sounds like using the same URL for all articles is probably a better approach. But I am not an expert in Facebook optimisation.

mikeller:
@blckmn: I grepped for references to `small_logo` in the repository and found none.

PR #60, opened 2020-08-25
Fixed mobile nav button change behavior

Asizon:
Fixed mobile nav button behavior when any nav menu item is selected.


Asizon:
Current problem screenshot:
Screenshot_20200825_110627_com android chrome

mikeller (approved):

PR #68, opened 2021-01-11
Adding in default region, and splitting build to only deploy on master.

blckmn:


Asizon:
No time to review before merged @mikeller?

PR #79, opened 2022-09-19
Replacing beginner video with Limon's

kensherman:
Replaces bardwells 4.3 betaflight series of videos with Limons

### DIFF
--- a/src/pages/index.ejs
+++ b/src/pages/index.ejs
@@ -203,7 +203,9 @@ description: "Betaflight is flight controller software (firmware) used to fly mu
 							<div class="container">
 								<div class="video-content">
 			
-									<iframe width="560" height="315"  src="https://www.youtube.com/embed/videoseries?list=PLwoDb7WF6c8nT4jjsE4VENEmwu9x8zDiE" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+									<div class="video-content">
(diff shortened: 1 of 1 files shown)


src/pages/index.ejs
```diff
@@ -203,7 +203,9 @@ description: "Betaflight is flight controller software (firmware) used to fly mu
 							<div class="container">
 								<div class="video-content">
 			
-									<iframe width="560" height="315"  src="https://www.youtube.com/embed/videoseries?list=PLwoDb7WF6c8nT4jjsE4VENEmwu9x8zDiE" title="YouTube video player" frameborder="0" allow="accelerometer; autoplay; clipboard-write; encrypted-media; gyroscope; picture-in-picture" allowfullscreen></iframe>
+									<div class="video-content">
```
Asizon, line 206: This div and class are already defined and opened

Asizon (changes requested):

src/pages/index.ejs
kensherman: The previous rev with the 4.2 videos used this presentation as it was a single video. When I put bardwells videos in they are a playlist which can only be embedded though an iframe. This help video is now back to a single video which I felt suited the oringal styling. My sense is either one works usability wise and I can go either way.

sugaarK (approved):

src/pages/index.ejs
Asizon: Good! But please remove this extra container, it already exist

sugaarK:
Ohh lols @Asizon  my bad

PR #80, opened 2022-09-22
Removing duplicated div

kensherman:


Asizon (approved):
Indendation is not ok at all, but aproved

src/pages/index.ejs
```diff
@@ -200,9 +200,7 @@ description: "Betaflight is flight controller software (firmware) used to fly mu
 						</div>
 						<section class="video-area col-lg-6">
 							<div class="overlay overlay-bg"></div>
-							<div class="container">
-								<div class="video-content">
-			
+								<div class="container">
```
haslinghuis, line 203:
```suggestion
						<div class="container">			
```

We also need to remove line 207

haslinghuis (changes requested):

src/pages/index.ejs
Asizon: Yeah good catch

sugaarK: @kensherman  fix this up and we’ll merge it

kensherman: Have run html checker on all files. There's a lot of noise from the esj script tags but think I've got the hmtl ones fixed

haslinghuis (dismissed):

blckmn (approved):

blckmn:
Needs a rebase

sugaarK (approved):

Asizon:
Ok, so this PR code breaks again our web page, @kensherman can you please fix it?

Asizon:
image

Asizon:
What is the reason to remove Header Id from code @kensherman ??

PR #88, opened 2023-01-13
Title case for consistent headers

VitroidFPV:
I think it's best we use [title case](https://www.computerhope.com/jargon/t/title-case.htm) for all headers going forward


haslinghuis:
function titleCase(str) {
  return str.toLowerCase().split(' ').map(function(word) {
    return (word.charAt(0).toUpperCase() + word.slice(1));
  }).join(' ');
}

PR #89, opened 2023-01-13
fix github deploy action so it includes CNAME to not break on deployment

freasy:
Before the CNAME file was mandatory for the custom domain setting for github pages.
Since i didnt know, i missed to copy it into the correct branch.

### DIFF
--- a/.github/workflows/docs.yaml
+++ b/.github/workflows/docs.yaml
@@ -6,6 +6,7 @@ on:
       - t-rex
       - master
       - develop
+      - fix-actions
     # Review gh actions docs if you want to further define triggers, paths, etc
     # https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#on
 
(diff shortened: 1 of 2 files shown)


.github/workflows/docs.yaml
```diff
@@ -24,12 +25,15 @@ jobs:
       - name: output
         id: output
         run: |
-          if [ ${{ github.repository_owner }} != 'betaflight' ]; then
-            echo "URL=https://${{ github.repository_owner }}.github.io" >> $GITHUB_OUTPUT
-            echo "BASE_PATH=/${GITHUB_REPOSITORY#*/}/" >> $GITHUB_OUTPUT
-          else
+          if [ ${{ github.repository_owner }} == 'freasy' ]; then
```
haslinghuis, line 28: This will never evaluate as true :)

freasy: this is so i can test github pages on my own repository

haslinghuis (approved):

PR #105, opened 2023-01-19
action runs on PR, readme and example env

freasy:
This adds an action to have a PR check (needs repository setup).

Also, it adds an example env file and the README entry for that.


freasy:
The check fails because of Markdown formatting errors. So my action works :D

haslinghuis (approved):

PR #111, opened 2023-01-21
Power Tab fixes - sidebar position, wording

VitroidFPV:
Draft PR, please review and replace the present placeholder text


haslinghuis:
Reviewers. We need some info here:

https://github.com/betaflight/betaflight.com/pull/111/files#diff-bb1dae1315af6d16a7599a9b05f6a297b8dbdeac06ac70bf8b76b974478fd7ddR61-R67

freasy:
```markdown
### Divider / Multiplier Value

This defines how the value read by the ICs adc converts into a voltage value that makes sense to the pilot.
In easy terms this sets how the ratio difference between the two voltage divider resistors are used in the voltage calculation formular.

Example:
Scale: 110 = 10:1 voltage divider (10k:1k ohm resistors)
Divider: 10 = sets the direct resistor to resistor ratio
Multiplier: 1 = this is to finetune the outcome of the calculation, a value of 1 is "no correction"
```

Reference to the code:
```c++
STATIC_UNIT_TESTED uint16_t voltageAdcToVoltage(const uint16_t src, const voltageSensorADCConfig_t *config)
{
    // calculate battery voltage based on ADC reading
    // result is Vbatt in 0.01V steps. 3.3V = ADC Vref, 0xFFF = 12bit adc, 110 = 10:1 voltage divider (10k:1k) * 100 for 0.01V
    return ((((uint32_t)src * config->vbatscale * getVrefMv() / 10 + (0xFFF * 5)) / (0xFFF * config->vbatresdivval)) / config->vbatresdivmultiplier);
}
```

docs/configurator/power-tab.mdx
```diff
@@ -56,13 +58,15 @@ Settings for the voltage meter to measure the battery voltage correctly
 The scale of the voltage meter. This is used to calculate the correct voltage from the ADC value. If the battery 
 voltage is not displayed correctly, you can adjust this value up/down to fix it
 
-### Divider Value
+### Divider / Multiplier Value
 
-*PLEASE REVIEW, NOT SURE WHAT TO WRITE HERE*
+This defines how the value read by the ICs adc converts into a voltage value that makes sense to the pilot.
```
haslinghuis, line 63: ADC :)

VitroidFPV:
Line 29 still needs content, really not sure what it could be though

docs/configurator/power-tab.mdx
haslinghuis: Please also take care of the merge conflict :point_down:

haslinghuis (approved):

freasy (approved):
LGTM

@SupaflyFPV please verify that my explanation for the voltage stuff is sufficient for a person that's new to betaflight.

JBKingdon:
I'm not sure what to do about the voltage divider info. The way the code behaves is that both divider and multiplier act as dividers. The reason is either lost in time or it's just a mistake. I'd be inclined to remove at least multiplier. Divider sets the units of the scale and allows more precision. So for a resistor network of r1:r2 scale is ((r1+r2)/r2)*divider. If a divider of 10 meets all expected use cases we could remove that from the UI as well and just document that the scale is in units of 1/10ths. I don't know off the top of my head if 10 works for everyone or not. We need to be able to adjust scale with fairly fine precision because tolerances on the resistors in the divider will often be 5 or 10% so scale will usually need some tuning.

PR #123, opened 2023-01-25
Fix links

haslinghuis:
- [x] fix links after moving wiki to it's own folder including `Manufacturer` and `Development` docs.  
- [x] add headers to wiki archive
- [x] fix naming to use only hyphens in configurator wiki pages (the extension of `-tab` is hardcoded in configurator and underscores are converted to hyphen)
- [x] rename `ACRO TRAINER` to `ACRO_TRAINER`


blckmn (dismissed):

SupaflyFPV (approved):

freasy:
Hard links with included domain makes it so that a fork can't click on the links.
I would recommend making them relative, so they are always valid inside every fork.

freasy (approved):
LGTM

(read my above comment)

haslinghuis:
@freasy Somehow a relative path did not work for some links. Since we are going to normalize files in a subsequent PR I leave it as is.

PR #126, opened 2023-01-26
Update hardware spec

haslinghuis:
Adding rules:

image

### DIFF
--- a/docs/manufacturer/Manufacturer Design Guidelines.md
+++ b/docs/manufacturer/Manufacturer Design Guidelines.md
@@ -15,6 +15,7 @@ Version Change Register
 | Draft 0.7 | 17 November 2022 | Remove off-board hardware defines |
 | Draft 0.8 | 01 January 2023  | Update BARO and CC2500            |
 | Draft 0.9 | 14 January 2023  | Add FC LEDs                       |
+| Draft 1.0 | 26 January 2023  | Add Signal Rules                  |
 
 Thank you for considering or continuing your development of Betaflight capable flight control hardware.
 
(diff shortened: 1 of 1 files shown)


SupaflyFPV:
Have a few comments from Steve to add bear with me...

SupaflyFPV (approved):

sugaarK:
@SupaflyFPV  what did Steve have to say

sugaarK (dismissed):

SupaflyFPV:
> @SupaflyFPV what did Steve have to say

"To achieve rule 1 you should use a ground plane. Be very careful splitting that plane by running tracks on the same layer especially near the gyro as that will cause noise."

He also shared a link about ground planes and I thought I should properly absorb it before committing to this guide...

https://www.nwengineeringllc.com/article/should-you-split-ground-planes-in-mixed-signal-pcb-design.php

sugaarK:
Good stuff thanks @SupaflyFPV

haslinghuis:
Should I add a note

SupaflyFPV:
> Should I add a note

My feeling is its too complex an area to insert as a note and I'd rather not risk bad guidance at this stage...I think we can work on it later...

haslinghuis:
image

haslinghuis:
Ok removed
```
To achieve the first rule you should use a ground plane. Be very careful splitting that plane by running tracks on the same layer especially near the gyro as that will cause noise. [reference](https://www.nwengineeringllc.com/article/should-you-split-ground-planes-in-mixed-signal-pcb-design.php)<br /><br />


```

SupaflyFPV (approved):

blckmn (approved):

PR #143, opened 2023-02-02
Fix broken links

blckmn:
- probably could have simply escaped the brackets.


blckmn:
@haslinghuis @freasy I think we should only run the checks on the files that have been changed in a PR.

You can get that list from actions such as what I used in the unified targets repo:
https://github.com/betaflight/unified-targets/blob/master/.github/workflows/deploy.yaml

haslinghuis (approved):

freasy (approved):

PR #158, opened 2023-02-22
Fix link

haslinghuis:
Sometimes we need a root slash before the link and sometimes not?


sugaarK (approved):

freasy (approved):
That depends on the location, I recommend always do the full path from root. So it's always right.

PR #160, opened 2023-03-04
Fix Katex Stylesheet

KarateBrot:
Found a small bug in the setup of the Katex stylesheet.

### Before
grafik

### After
grafik

### DIFF
--- a/src/css/tailwind.scss
+++ b/src/css/tailwind.scss
@@ -9,10 +9,6 @@
 
 .theme-doc-markdown {
 
-  *[aria-hidden="false"] {
-	@apply hidden;
-  }
-
   h1,
   header h1 {
     @apply text-6xl font-bold text-primary-500;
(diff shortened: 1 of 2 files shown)


src/css/tailwind.scss
```diff
@@ -5,10 +5,11 @@
 .theme-doc-markdown a, .fancy-link {
     @apply no-underline bg-gradient-to-r from-primary-400 to-primary-400 bg-no-repeat bg-left-bottom 
 		bg-0w hover:bg-100w hover:text-primary-500 duration-300 w-fit text-primary-200 transition-all;
-  }
+}
+
 .theme-doc-markdown {
 
-  *[aria-hidden="true"] {
+  *[aria-hidden="false"] {
```
freasy, line 12: this should not be changed, maybe a debug statement?

if aria hidden is set to true, hidden shall be applied.

freasy (changes requested):

src/css/tailwind.scss
freasy: Yes remove the whole block.

haslinghuis (approved):

freasy (approved):

PR #226, opened 2023-09-12
Document dshot_telemetry_start_margin setting

SteveCEvans:
I considered making this setting automatic, but I think it's better that this be manually set as with good ESC firmware there shouldn't be an issue and it would be bad to mask bad behaviour which should be fixed.


SteveCEvans:
After some discussion, making this automatic would be better, so closing this PR.

PR #242, opened 2023-11-14
Fix wording in introduction.mdx

valeriyvan:


nerdCopter (approved):

nerdCopter:
thank you.
`Precision` also acceptable -- maybe other devs will comment a preference.

haslinghuis (approved):
From https://wikidiff.com/precise/precision:

As adjectives the difference between precise and precision is that precise is exact, accurate while precision is used for exact or precise measurement.

As a noun precision is the state of being precise or exact; exactness.

SupaflyFPV (commented):
Thanks for the PR correction - I meant precision. I used this term to describe the system of tuning we have in Betaflight which uses blackbox logging and associated tools - ie evidence based - which provides the user with precision. 
The result is a precise tuning - so either can be used. I prefer precision because it refers to the evidence based methodology in Betaflight compared to other firmwares.

haslinghuis (changes requested):

haslinghuis (approved):

nerdCopter (approved):

PR #244, opened 2023-11-18
Calibration notes update

ctzsnooze:
Minor update of the mag calibration process to align with the recently merged [PR 12998](https://github.com/betaflight/betaflight/pull/12998)

In view of the dubious nature of the old driver code for the AK8963 and AK8975 (see comments from ledvinap on the mag scheduler firmware PR), I added a 'caution' flag indicating that these drivers were deprecated and that they may not work or may cause other i2c issues.  

At this point the only mag with exhaustive testing is the QMC5883L.  Others should work but no-one seems to have time to test them.

### DIFF
--- a/docs/wiki/archive/Magnetometer.md
+++ b/docs/wiki/archive/Magnetometer.md
@@ -93,7 +93,7 @@ In Betaflight 4.5, the CLI variable `mag_declination` was introduced, to correct
 
 Flight / FC firmware must be specially built to include `Magnetometers` for the cloud build, or `-DUSE_MAG` for local builds, otherwise there will be no Mag support in the firmware on the FC. Additionally, GPS firmware should be included in the build, because we display the Mag heading on Configurator's GPS tab, and use GPS debugs to display the Mag heading information.
 
-Betaflight provides drivers the following magnetometers, but not all have been validated to work with Betaflight 4.5's revised scheduling code:
+Betaflight provides drivers for the following magnetometers, but not all have been validated to work with Betaflight 4.5's revised scheduling code:
 
 - [QMC5883L](https://datasheet.lcsc.com/szlcsc/QST-QMC5883L-TR_C192585.pdf)[^8]. The QMC5883L is provided on the common, and cheap, GY-217 module, and many GPS units. It provides a 200Hz data update rate, 8x sample averaging and 3000 LSB/Gauss sensitivity. Standard axis orientation. We recommend using this mag if it is an option for your build, because it's performance has been carefully validated during testing and we know it works well.
 - [IST8310](https://intofpv.com/attachment.php?aid=8104)[^9] Note that this gyro has a highly unusual axis orientation, with Y to the _right_ when X is forward and Z is up, and will _always_ require a custom axis orientation in the CLI. Data update rate is 160Hz with 16x sample averaging and 330 LSB/Gauss sensitivity.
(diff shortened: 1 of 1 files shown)


docs/wiki/archive/Magnetometer.md
```diff
@@ -261,57 +260,53 @@ Calibration values, once acquired, typically do not need to be changed much with
...
+The accuracy of the calibration values can be validated by checking the absolute minimum and maximum values, for each axis, using Configurator's Sensors tab. With a 'perfect' calibration, the min and max for any individual axis, at the point when the values on the other two axes are zero, should be numerically equal.
 
 Using the sensors tab in this manner to check max and min for each axis can also validate the accuracy of the existing calibration when flying in a new location. If max and min are very similar, you don't need to recalibrate.
 
-This method also confirms that the orientation of the Mag is correct, by confirming max positive:
+Checking in Sensors confirms that the orientation of the Mag is correct, by confirming that the maximum positive values are as follows:
 
-- on X, when the nose points into the field,
+- on X, when the nose points durectly into the field (ie Nose points to Magnetic North at the local field inclination angle),
```
haslinghuis, line 309:
```suggestion
- on X, when the nose points directly into the field (ie Nose points to Magnetic North at the local field inclination angle),
```

docs/wiki/archive/Magnetometer.md
```diff
@@ -261,57 +260,53 @@ Calibration values, once acquired, typically do not need to be changed much with
...
+It's also possible to do a 'manual' calibration 'one axis at a time'. For example, if we want to get an accurate cal value for the X axis, we can point the nose randomly in the general direction of Magnetic field North (which requires knowing both the declination and inclination angles) for some 12-13s, then rotate the quad 180 degrees, and randomly point the tail into the field for the same period of time. Alternatively, we can make full 360 degree turns with the nose pointing generally in the North-South field plane. This will _only_ return an accurate Cal value for the X axis (the values for the other axes will be incorect). You can get then the cal value for X from the CLI and write it down. After repeating the same process separately for Y and Z, you will have all three cal values. Then you can type in the cal value for each axis into the CLI.
```
haslinghuis, line 295:
```suggestion
It's also possible to do a 'manual' calibration 'one axis at a time'. For example, if we want to get an accurate cal value for the X axis, we can point the nose randomly in the general direction of Magnetic field North (which requires knowing both the declination and inclination angles) for some 12-13s, then rotate the quad 180 degrees, and randomly point the tail into the field for the same period of time. Alternatively, we can make full 360 degree turns with the nose pointing generally in the North-South field plane. This will _only_ return an accurate Cal value for the X axis (the values for the other axes will be incorrect). You can get then the cal value for X from the CLI and write it down. After repeating the same process separately for Y and Z, you will have all three cal values. Then you can type in the cal value for each axis into the CLI.
```

ctzsnooze: Thanks, Mark.  Updated and slightly revised.

docs/wiki/archive/Magnetometer.md
ctzsnooze: Thanks again for checking this.

docs/wiki/archive/Magnetometer.md
```diff
@@ -93,13 +93,17 @@ In Betaflight 4.5, the CLI variable `mag_declination` was introduced, to correct
 
 Flight / FC firmware must be specially built to include `Magnetometers` for the cloud build, or `-DUSE_MAG` for local builds, otherwise there will be no Mag support in the firmware on the FC. Additionally, GPS firmware should be included in the build, because we display the Mag heading on Configurator's GPS tab, and use GPS debugs to display the Mag heading information.
 
-Betaflight supports the following magnetometers:
+Betaflight provides drivers the following magnetometers, but not all have been validated to work with Betaflight 4.5's revised scheduling code:
```
haslinghuis, line 96:
```suggestion
Betaflight provides drivers for the following magnetometers, but not all have been validated to work with Betaflight 4.5's revised scheduling code:
```

haslinghuis (approved):

PR #260, opened 2023-12-11
Update Nodejs to 20.10

haslinghuis:


ctzsnooze:
Let's move to node 20 only when each of Configurator, Betaflight.com, and Blackbox are all compatible.  
Otherwise chopping and changing node versions is a real headache.

haslinghuis:
This will make it the same as blackbox-log-viewer. App-dmg is main issue with configurator going forward.

nerdCopter:
"20.x" is considered the current "LTS" at this point: https://nodejs.org/en/about/previous-releases.
My opinion is that with **_thorough_** testing it can be allowed, but the new/different OSX chips will be the determinant factor i believe.

KarateBrot (approved):

nerdCopter (approved):
* builds/runs successfully locally Debian 11. `nvm use 20` ;  npm install ; cp -p .env.example .env ; npm start`
* has stale docs, but presume merge will be fine.

PR #267, opened 2023-12-13
Rename wiki to support

haslinghuis:


nerdCopter (approved):
* approving without deep inspecting.
* i still believe a wiki>support redirect would be useful for any pilot's bookmarks, posted links, etc.

haslinghuis:
We introduced this site not long ago, so most links [based on old wiki] would be invalid.
A lot of work has still to be done to update the site.

VitroidFPV (approved):

PR #281, opened 2023-12-18
Give getting started a default index page, and update it a little

ctzsnooze:
Hi guys

>after recent merges, all we need is to change the index document attributes to avoid the duplication of the file name.

original comment:

It loads 'index.md' (which is the old Getting Started document, re-named) when the user clicks on the `Getting Started` sidebar element.  Instead of just showing the contents of the directory, which we already see below `Getting Started` anyway.

The other three documents in the directory appear under the heading.  

Hence each sidebar element has its own unique page, and the page to the right of a sidebar header also can be more useful.  I've done something similar with the Guides section in a recent PR.  To avoid re-duplicating the main page name at the top, I gave it an ID and a title.

Not sure if this is the recommended way, but it works :-)  

Looks like this, after clicking on `Getting Started`:

Screen Shot 2023-12-18 at 16 49 06

### DIFF
--- a/docs/wiki/getting-started/index.md
+++ b/docs/wiki/getting-started/index.md
@@ -1,15 +1,13 @@
 ---
 sidebar_position: 0
+id: Getting started
+title: About Betaflight
 ---
 
-# Introduction to Betaflight
+Betaflight is flight controller software (firmware) for multi-rotor and fixed wing craft.
 
-## What is Betaflight?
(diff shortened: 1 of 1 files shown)


docs/wiki/getting-started/index.md
```diff
@@ -0,0 +1,101 @@
+---
+sidebar_position: 0
+id: Getting started
+title: About Betaflight
+---
+
+Betaflight is flight controller software (firmware) for fly multi-rotor fixed wing craft.
```
nerdCopter, line 7:
```suggestion
Betaflight is flight controller software (firmware) for flying multi-rotor fixed wing craft.
```
or `controlling` or `piloting` :man_shrugging:

nerdCopter:
in light mode, graphing outlines are greenish.
image

ctzsnooze:
@nerdCopter this PR is not consideringthe graphing outlines, the graphing is handled by @VitroidFPV, has to be fixed in a different PR.

haslinghuis (approved):

nerdCopter (approved):

PR #294, opened 2023-12-21
Style: Change doc card spacing and layout

VitroidFPV:
image

### DIFF
--- a/src/css/tailwind.scss
+++ b/src/css/tailwind.scss
@@ -199,6 +199,18 @@ nav.menu.thin-scrollbar {
   }
 }
 
+.cardContainer_src-theme-DocCard-styles-module {
+	@apply border-0 border-l-4 border-transparent p-3 duration-300 #{!important};
+}
+
+.cardContainer_src-theme-DocCard-styles-module:hover {
+	@apply border-primary-500 bg-primary-500/20 #{!important};
+}
+
+.cardTitle_src-theme-DocCard-styles-module {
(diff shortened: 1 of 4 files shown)


VitroidFPV:
Pull and try it locally, I'm working on a laptop so some things may look different than on regular sized screens

ctzsnooze:
Thanks - wow much more complicated than I had hoped.  My apologies that it required so much work!

Looks great!

On a phone, we get about two more labels per screen than before, and it looks good, something approximately like this (I removed the `learn all of the configurator tabs and options` line):

Screen Shot 2023-12-22 at 10 35 17

In most compact 2-column mode I personally thing there is too much padding around the text, that's just me perhaps, but it's no big thing, I'm happy with it the way it is.

Screen Shot 2023-12-22 at 10 37 52

In a big screen, we don't get the 'headings' or 'contents' menu on the right, so we get a kind of waste of space there, but it looks great!

Screen Shot 2023-12-22 at 10 40 45

Definite improvement over master, so I'd be happy to merge this.

Much respect @VitroidFPV that's a ton of work, very specialised.  Thank you!

ctzsnooze (approved):

ctzsnooze:
One small detail, is there any way we can omit the 'contents' space on the right when the page has no 'contents' to display?  
In list-only directories like say 'boards' or 'guides' it looks a bit strange to have that empty space there on the right.

sugaarK (approved):

PR #305, opened 2023-12-23
Content: Add FC doc template and example

VitroidFPV:
Feedback is strongly encouraged! If this is to be the future format for all the board docs, we should make it as good as possible


nerdCopter:
```
Compiled with problems:
ERROR in ./docs/development/manufacturer/fc_documentation/fc-doc-template.mdx
Module build failed (from ./node_modules/@docusaurus/mdx-loader/lib/index.js):
Error: MDX compilation failed for file "/......../betaflight.com/docs/development/manufacturer/fc_documentation/fc-doc-template.mdx"
Cause: Could not parse expression with acorn
```

nerdCopter:
image

ctzsnooze:
Fundamentally these documents are for users, so need to be user-focused.

I think that for consistency in the "Main Features: area we should list all features that may apply for all builds, in a consistent order, with the manufacturer indicating "None" when something isn't there.  Otherwise the user may wonder if, for example, flash memory is there or not, when it is not listed; if, on the other hand, it says Blackbox : none, or Barometer : none, it's very clear.

Likewise for connectors, list all the kinds of connectors commonly used, and say 'none' if none for that type, For instance, USB (type), ESC (type/none, pin sequence), HD Video (type/none, pin sequence), Serial (type/none, number).

The pad listing template is really good, again requiring 'none' if something is not present.

So that photos associate with the board, my preference would be a directory, at same level as the board, named `BOARDNAME_IMAGES` or something like that.  Easy to find the matching images.  Otherwise in static/img/boards/BOARDNAME.

haslinghuis:
Please let's finish this so we can add this to the requirements for manufacturers when adding new targets.

nerdCopter:
@VitroidFPV , is the template supposed to render properly, or is it just supposed to be a downloadable template text?
if the latter, maybe the entire template should be in a 3-backtick block, or maybe just make the template a downloadable file rather than rendering it at all. i'm unsure the best method, but the rendering of the template does not currently include the data ion the tables.

ctzsnooze (approved):

haslinghuis (approved):

PR #310, opened 2023-12-29
Javascript re-directs for possible user URLs

ctzsnooze:
This redirect method uses a javascript method recommended by openAi.

The intent is that certain 'common' user links can be entered in the form `https://betaflight.com/nnn` and not get a 404 page not found error.  So we can tell people to go to the wiki on betaflight.com and if they type `https://betaflight.com/wiki`, they will get to our wiki landing page.  Whereas without this PR they will get a 404 page not found error.

The coding for these redirects works properly with `yarn start` and also works when I make the full html build version locally using `yarn build`.  

We are currently using a method on the sponsors.mdx page where a 'page' is put in `src/pages/sponsors.mdx` with code that replaces it with the current `/docs/relationships/sponsors.mdx` page when the static site is created. 

This works, though with issues, and I don't want to break what we have, so let's leave that alone for now..

The outcome should be as per this table - we could add more if desired:

| URL    | returns |
| -------- | ------- |
| https://betaflight.com/about | https://betaflight.com/docs/wiki/getting-started |
| https://betaflight.com/development | https://betaflight.com/docs/development     |
| https://betaflight.com/guides  | https://betaflight.com/docs/wiki/guides    |
| https://betaflight.com/releases  | https://betaflight.com/docs/category/release-notes    |
| https://betaflight.com/setup  | https://betaflight.com/docs/wiki/getting-started/setup-guide    |
| https://betaflight.com/sponsors  | not changed from current method   |
| https://betaflight.com/wiki  | https://betaflight.com/docs/wiki   |

So far I've had no concrete suggestions for alternative methods.  I would be grateful if this could be merged.  So far no-one has coded and tested an alternate proposal.  This one works well in both static and dynamic site build tests.  Please let's get the functionality up there.  We can always change the mechanism used if this proves problematic or a better solution can be found.


nerdCopter:
works as intended with the following warnings:
```
[WARNING] Duplicate routes found!
- Attempting to create page at /about, but a page already exists at this route.
- Attempting to create page at /development, but a page already exists at this route.
- Attempting to create page at /guides, but a page already exists at this route.
- Attempting to create page at /releases, but a page already exists at this route.
- Attempting to create page at /setup, but a page already exists at this route.
- Attempting to create page at /sponsors, but a page already exists at this route.
- Attempting to create page at /wiki, but a page already exists at this route.
This could lead to non-deterministic routing behavior.
```

should aliases be a function of the server, not the code?

KarateBrot:
There is a lot of code duplication here. Instead of so many duplications, we could pass the link into the method as an argument, so we only need to define it once.

haslinghuis:
Probably won't work with Github Pages as it's static.

This is looks like an AI proposal. Don't think users are going to request URL's like this. We already have a search option and we did not need this redirection before.

ctzsnooze:
@KarateBrot all client side redirects require a static page to redirect from, and the javascript has to be in a separate file.  If all the javascript could be in one file, that would be nice, and if there was a tidier method, that would be great, however no-one has yet proposed a better alternative.

@haslinghuis We already require redirects for the `betaflight.com/sponsors` URL.  Redirects for common terms are very helpful.  Its nice to be able to say to someone, go to `betaflight.com/help` or `betaflight.com/manufacturers` and have them end up somewhere useful - wherever that landing page happens to be.  I think this will be a very powerful improvement in usability.  If we had redirects for most common user searches, the user experience will be easier, and most 'landing pages' can be directly accessed.

It also helps us.  For example, by redirecting `betaflight.com/sponsors` to the `sponsors` landing page, that link remains intact even if, for some internal reason, we move the actual location of the page from where it is now, to somewhere else.

We can then use the redirected link internally also without having to update it all the time.

For instance, a link to the 'latest release information' can use `betaflight.com/release` and we only have to change the redirect to point to wherever the latest release info is.

I think this will be very, very useful.

blckmn:
We definitely need simplified URLs.

betaflight.com/sponsors
betaflight.com/wiki
betaflight.com/releases

etc. Easy to remember, and easy to update content or underlying structure without having to go around and cleanup links. This is especially important as the configurator is released, and we don't want links changing between versions - they are immutable.

Github pages is static. So this static content will ultimately be duplicated when produced into the "static" representation, however it should only be defined once in the repo, it is just "prepared" twice as needed.

nerdCopter:
i found some options, but did not dig in nor deeply review:
- https://docs.github.com/en/contributing/writing-for-github-docs/configuring-redirects
- https://lea.verou.me/blog/2016/11/url-rewriting-with-github-pages/
- https://theorangeone.net/posts/redirecting-static-pages/

ctzsnooze:
@blckmn @nerdCopter @haslinghuis 
I've changed the PR to use the current method for sponsors.
This puts a page `sponsors.mdx` in the `/src/pages/` directory, which contains code replacing the contents with the current version of /docs/relations/sponsors.mdx`.
We know this works for sponsors.

nerdCopter:
@VitroidFPV @freasy

VitroidFPV (approved):

haslinghuis (approved):

PR #327, opened 2024-01-09
Update mag i2c and IST8310 notes

ctzsnooze:
The IST8310 Mag can have any one of four 12c addresses. The default is 12, but it could also be 13, 14 or 15.  At least one integrated GPS module, the GEP-M1025-MI, uses 14.

This PR clarifies i2c a bit better, and provides information about how to handle Mag units that are not detected, as well as specific notes about the IST8310.

Sure would be easier to explain i2c if we used the term 'bus' instead of 'device' when we refer to the Flight Controller's physical i2c bus to which Mag or Baro devices are connected.

### DIFF
--- a/docs/wiki/guides/current/Magnetometer.md
+++ b/docs/wiki/guides/current/Magnetometer.md
@@ -104,12 +104,14 @@ In Betaflight 4.5, the CLI variable `mag_declination` was introduced, to correct
 
 ## Hardware and Connection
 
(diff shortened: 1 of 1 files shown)


nerdCopter (approved):

haslinghuis (approved):

haslinghuis:
Think terminological is correct. We configure the `mag_device` address as we configure bus type with `mag_bustype`

PR #334, opened 2024-01-13
Add note about BMP-270 gyro, DPS310 clone, IST8310 mag, to manufacturer guidelines

ctzsnooze:
Wasn't sure where to put this, but under IMU seems best, without having to re-order all the subheadings.

The intent is to:

- recommend against using the Bosch BMI-270 gyro, which has been identified as having un-calibrated gyro data.  When the quad is turning more than 15 deg/s, integrated gyro data is used to determine the change in attitude of the quad.  Only when the rate of change is less than 15 deg/s will the accelerometer data be used.  Integration of the BMI-270 leads to up to 5% errors in angle estimate, and the result is that after a quick input, there will be a kind of 'jiggle' when the IMU shifts orientation back to the accelerometer value.  This is highly undesirable.  Hence I suggest that we strongly recommend against using the BMI-270

- recommend against using barometer clones, and strongly recommend the 'real' Infineon DPS368, which replaces the DPS310.  Infineon suggest accuracy within 10cm despite temperature change.  On testing, only the 'real' infineon chips report correct temperature adjustment values, and the clones are more likely to be associated with incorrect altitude readings.

- recommend against the IST8310 mag, which has the Y axis reversed from the normal position, and consequently can only be 'custom' oriented.  Simple rotations and inversions which work for orientation corrections of the QMC5883L do not work with the IST8310.  It will always require a custom correction.

- recommend using the default i2c addresses for barometer and magnetometer chips, so that we will auto-detect correctly.

If this information should be somewhere else, or somewhere else as well, let me know.

PS: I had to set the file extension to .md, not .mdx, because I cannot commit a file with the the extension .mdx without a lint and husky error that provides inadequate information for me to resolve.  If someone can help fix my local build environment so I don't get that error, and can build with .mdx file extensions, that would be great.

### DIFF
--- a/docs/development/manufacturer/manufacturer-design-guidelines.md
+++ b/docs/development/manufacturer/manufacturer-design-guidelines.md
@@ -203,6 +203,7 @@ The IST8310 magnetometer has an unusual axis orientation, where the Y axis is 18
 The use of magnetometers with non-standard axis orientations is not recommended.
 
 :::
+
 Note also that the IST8310 magnetometer can be configured with any one of four i2C adresses. Betaflight will only connect to the IST8310 automatically if the default i2C address of 0x0C is used. If any of the other three i2C addresses (0x0D, 0x0E, 0x0F) are used, the user will need to custom enter either 13, 14 or 15 as the `mag_i2c_address` value, or it will not work.
 
 The QMC5883L has 'normal' axis orientation and works well.
--- a/docs/development/manufacturer/manufacturer-design-guidelines.md
+++ b/docs/development/manufacturer/manufacturer-design-guidelines.md
@@ -29,6 +29,7 @@ import useBaseUrl from '@docusaurus/useBaseUrl'
 | Draft 0.9 | 14 January 2023  | Add FC LEDs                       |
 | Draft 1.0 | 26 January 2023  | Add Signal Rules                  |
 | Draft 1.1 | 10 December 2023 | Add LSM6DSV16X and LPS22DF        |
+| Draft 1.2 | 13 January 2024  | Add Mag and Baro hardware note    |
 
 Thank you for considering or continuing your development of Betaflight capable flight control hardware.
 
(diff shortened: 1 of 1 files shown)


docs/development/manufacturer/manufacturer-design-guidelines.md
```diff
@@ -165,6 +166,53 @@ For connector pinout please refer to the [Betaflight Connector Standard](connect
 
 Selecting an appropriate IMU for flight controller operation is critical to the resulting flight performance of systems. Proven examples of hardware using single Invensense MPU-6000, single or dual ICM-20602, and also Bosch BMI-270 and BMI-180 units have been successfully demonstrated to operate with Betaflight, although the latter two examples have required significand development effort to bring performance of the IMU gyroscope and accelerometer sensing behaviors up to the standards required to maximize flight performance.
 
+:::note
```
haslinghuis, line 169: Admonitions won't work with .md extension?

haslinghuis: @VitroidFPV @freasy please look into the fact we are not able to commit when having set up local build with npm while changing files with mdx extension.

Have used two terminals on same PR in different folders - both with the changes wanting to submit - but only one with build setup for testing. Using first terminal to commit.

This seems to work as git clean -xdf does not remove everything after having build setup.

ctzsnooze: Yes I have tried all manner of things and the only way I can commit a file with an .mdx extension is to block all checking.

ctzsnooze: `:::note` `:::warning` and similar admonitions all work in files with an `.md` extension.  I think docusaurus handles them while creating the html, regardless of the extension.

ctzsnooze: I think this is the right place, it is information for manufacturers.

haslinghuis: @ctzsnooze Did not verify 

@VitroidFPV @freasy but would it mean we don't have to use mdx extension at all? Simplifying contribution here?

Single document changes can be made easy with github functionality. (Maybe even have a button `edit this page on github` to stimulate contribution).

VitroidFPV: I will have a fix PR up hopefully by today

limonspb:
Where would I read more about BMI 270 problem?
I think I remember inav had problems with BMI and angle + modes, but somehow fixed them, or tried

ctzsnooze:
The BMI issue was noted in [issue 11894](https://github.com/betaflight/betaflight/issues/11894), with the video [here](https://www.youtube.com/watch?v=hnNc33QCbM8).

It was discussed in detail in discord in the public development channel around 6th Jan 2024.

I made some videos.  In 4.5 the behaviour is easy to replicate, but the test method from the youtube video doesn't work, because we now more rapidly correct than before, transiently, when a rapid movement stops.  Hence I'm planning to make a PR that would allow better testing.

docs/development/manufacturer/manufacturer-design-guidelines.md
```diff
@@ -165,6 +166,53 @@ For connector pinout please refer to the [Betaflight Connector Standard](connect
...
+
+**Magnetometer selection**
+
+The IST8310 magnetometer has an unusual axis orientation, where the Y axis is 180 degrees rotated compared to all the other commonly used magnetometers. The user must use custom mag orientation values for correct functioning of the IST8310. It is not possible for the user to just apply a simple rotation to correct the IST8310. For this reason we do not recommend using the IST8310 magnetometer with Betaflight.
+
+:::note
+
+The use of magnetometers with non-standard axis orientations is not recommended.
+
+:::
```
haslinghuis, line 205:
```suggestion
:::

```

haslinghuis (approved):

docs/development/manufacturer/manufacturer-design-guidelines.md
nerdCopter: anything else to do? or is this approve-able as is?

ctzsnooze: I think it can be approved and merged as-is.  There appears to be no functional difference in pages whether their extension is .md or .mdx.  We are hoping that @VitroidFPV can figure out why I can commit a change for .md files but not .mdx, that's all.

nerdCopter (approved):

bw1129:
I performed the same test on a RADIX2 HD (with bmi270), and I do NOT see any issue 
[360 test on RADIX2 HD](https://youtu.be/RNHdr__mB-U?si=aU1ngfYYAOLAqQ0q)

Is this the only source for this claim? 
[original vid illustrating issue](https://youtu.be/hnNc33QCbM8?si=-vyP7b89JfPmJdnK)

bw1129:
because the issue appeared to compound with number of rotations in same direction, I ran the test again but I rotated multiple times in same direction, then back. Same results. error <= 0.2 deg. No issue
[test 2 multiple rotations](https://youtu.be/ek4-GwXpvvI?si=MMJt8kF498Qbwh9E)

mluessi:
@bw1129 it doesn't happen with the RADIX 2 HD because the BMI270 has been calibrated.

nerdCopter:
can an end-user (pilot) calibrate? or is it micro-code based?

PR #340, opened 2024-01-18
Update Barometer wiki page

ctzsnooze:
Tried to update the Barometer page to:

- provide general guidance about using a Baro
- help solve typical 4.5 issues where the firmware didn't support a specific baro
- explain how to configure an external baro.

Moved the old document to archive

Added a shortcut so that betaflight.com/baro goes straight to the betaflight.com/wiki/guides/current/barometer page.


VitroidFPV (approved):

SteveCEvans:
Maybe also mention `set altitude_source = BARO_ONLY` to ensure that the baro reading is used in preference to the less accurate GPS altitude.

haslinghuis (approved):

ctzsnooze:
Good point, @SteveCEvans. 
Quite difficult to explain that stuff, but hopefully the most recent commit helps a little.

PR #347, opened 2024-01-24
Update 4.5 release notes with OSD HD default and others

ctzsnooze:
Some quick notes about the default behaviour when OSD_HD is included in the build (clean flash will configure the board for OSD_HD, not for Analog, by default).

A few other tidy-ups.

Anything else we should mention / do?

### DIFF
--- a/docs/wiki/release/Betaflight-4-5-Release-Notes.md
+++ b/docs/wiki/release/Betaflight-4-5-Release-Notes.md
@@ -10,7 +10,9 @@ Betaflight 4.5 is an incremental release. The basic flight parameters have not c
 IMPORTANT: use Configurator 10.10!  The most recent release version is available [here](https://github.com/betaflight/betaflight-configurator/tags), or use the [latest nightly build](https://github.com/betaflight/betaflight-configurator-nightlies).
 :::
 
-As usual, Full Chip Erase is mandatory when flashing. Re-configuring from scratch is safer than importing a CLI dump or a saved Preset.  Users of GPS Rescue, Angle and Horizon modes shold NOT use their old values.  Otherwise, most flight, Rx, Mode, OSD, and GPS parameters have not changed since 4.4.  Any new or re-named parameters will get default values, so importing a 4.4 save file (Presets>Save) is in most cases, OK.
+As usual, Full Chip Erase is mandatory when flashing. Re-configuring from scratch is safer than importing a CLI dump or a saved Preset.  Users of GPS Rescue, Angle and Horizon modes shold NOT use their old values.  Otherwise, most flight, Rx, Mode, OSD, and GPS parameters have not changed since 4.4.  Any new or re-named parameters will get default values, so importing a 4.4 save file (Presets>Save) is, in most cases, OK.
+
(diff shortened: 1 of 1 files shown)


docs/wiki/release/Betaflight-4-5-Release-Notes.md
```diff
@@ -10,7 +10,9 @@ Betaflight 4.5 is an incremental release. The basic flight parameters have not c
...
 :::
 
-As usual, Full Chip Erase is mandatory when flashing. Re-configuring from scratch is safer than importing a CLI dump or a saved Preset.  Users of GPS Rescue, Angle and Horizon modes shold NOT use their old values.  Otherwise, most flight, Rx, Mode, OSD, and GPS parameters have not changed since 4.4.  Any new or re-named parameters will get default values, so importing a 4.4 save file (Presets>Save) is in most cases, OK.
+As usual, Full Chip Erase is mandatory when flashing. Re-configuring from scratch is safer than importing a CLI dump or a saved Preset.  Users of GPS Rescue, Angle and Horizon modes shold NOT use their old values.  Otherwise, most flight, Rx, Mode, OSD, and GPS parameters have not changed since 4.4.  Any new or re-named parameters will get default values, so importing a 4.4 save file (Presets>Save) is, in most cases, OK.
+
+If the `HD_OSD` option is included in the build, after a clean flash, all relevant OSD setups values will suit HD video.  Analog users should visit Configurator's OSD tab and change the Video Format to either Auto, NTSC or PAL, and save.  If only `SD_OSD` is included in the build, a clean flash will default to Analog OSD settings.
```
haslinghuis, line 15:
```suggestion
If the `OSD_HD` option is included in the build, after a clean flash, all relevant OSD setups values will suit HD video.  Analog users should visit Configurator's OSD tab and change the Video Format to either Auto, NTSC or PAL, and save.  If only `OSD_SD` is included in the build, a clean flash will default to Analog OSD settings.
```

Please also recommend to either use OSD_SD or OSD_HD depending on hardware as switching does not work correctly (midcol setting).

ctzsnooze: Thanks, that was silly :-)
I think that people will only swap from Analog to HD, after a clean flash, right at the start.  After a clean flash, no elements will be configured.  Your firmware PR should set the initial element correctly now.  
Once setup, a person isn't likely to swap Analog to HD and back all the time just for fun.  Typically they will set it up for the camera, or the format, and leave it there.  They have to move things around anyway, so I don't think we need to worry too much about that. The need to manually re-position has always been like that, anyway, it's not a new 4.5 feature :-)
A lot of users will apply HD or SD Presets with the Preset's elements positioned to suit, then they don't need to worry about it at all, other than re-arranging.
Hence I don't think we need to do auto-repositioning when swapping modes, and I think that by making the clean default 'right', that's OK.

haslinghuis (approved):

nerdCopter (dismissed):

nerdCopter (approved):

haslinghuis (approved):

PR #356, opened 2024-01-30
Remove attempt to use embedded HTML for image

howels:
Removing the embedded HTML attempt to control the image size on `dshot_bit.webp`

If anyone knows how to limit the size of images and prevent them scaling to full-width please let me know.

### DIFF
--- a/docs/development/API/Dshot.md
+++ b/docs/development/API/Dshot.md
@@ -41,7 +41,6 @@ SSSSSSSSSSSTCCCC
 
 The interesting part is, that 1 and 0 in the Dshot frame are distinguished by their high time. This means that every bit has a certain (constant) length, and the length of the high part of the bit dictates if a 1 or 0 is being received.
 
-<img src="dshot_bit.webp" alt="Dshot high times" width="400"/>
 ![Dshot high times](./dshot_bit.webp)
 
 This has two benefits:


nerdCopter:
@howels , found a solution that works.  resize the .webp file to 200px wdth, 168px height.
attaching image for re-use - [dshot_bit.webp.zip](https://github.com/betaflight/betaflight.com/files/14101361/dshot_bit.webp.zip)


rendered:
image
image

haslinghuis:
Should be easy with tailwind css but cannot get it to work:

```
<div class="w-1/2">
    <img class="h-auto max-w-xs" src="/img/dshot_bit.webp" alt="image description"/>
</div>
```

howels:
> Should be easy with tailwind css but cannot get it to work:
> 
> ```
> <div class="w-1/2">
>     image description
> </div>
> ```

Do we simply resize the images then? Would be nice if we could say "50% width" instead of using a fixed size.

nerdCopter:
> [...] Would be nice if we could say "50% width" instead of using a fixed size.

agree, but since attempts failed.  physical scale worked as quick-fix. maybe it is a GitHub-only issue as some resources allude.

i think this PR can be closed as https://github.com/betaflight/betaflight.com/pull/357 contains the quick-fix.

howels:
> > [...] Would be nice if we could say "50% width" instead of using a fixed size.
> 
> agree, but since attempts have failed, physical scaling worked as quick-fix. maybe it is a GitHub-only issue as some resources allude.
> 
> i think this PR can be closed as #357 contains the quick-fix.

This PR removes a line of redundant HTML code that I left in by accident

PR #357, opened 2024-01-30
Update Dshot.md

ot0tot:
Cleanup formatting and typographical errors.

note: ot0tot = V-22

PR #361, opened 2024-01-31
Update Dshot.md

ot0tot:
Resolve remaining comments from PR #357. Update discussion on relationship between frame rate and PID loop rate.

### DIFF
--- a/docs/development/API/Dshot.md
+++ b/docs/development/API/Dshot.md
@@ -23,9 +23,9 @@ But, everything has two sides, and so do digital protocols. The downsides are, t
 
 ## Supported Hardware
 
-DShot is supported on all BLHELI_S, BLEHLI_32 and KISS ESC’s. One limitation are older BLHELI_S ESC’s with EFM8BB1 MCU: Only DShot 150 and DShot 300 are supported on those, but this should still be good enough for 99% of use cases.
+DShot is supported by all BLHELI_S, BLHELI_32 and KISS ESCs. Note that older BLHELI_S ESC’s with EFM8BB1 MCUs only support DShot 150 and DShot 300, but this should be fine for most cases.
 
-No extra settings on the ESC’s are needed - they automatically detect by which protocol they are driven and act accordingly. Some firmware might only support a certain subset of protocols though, so be aware of that. Bluejay for example only supports DShot in all its variations, but none of the analog protocols.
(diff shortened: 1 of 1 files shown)


ot0tot:
Marked as draft until stylesuxx has a chance to provide feedback.

docs/development/API/Dshot.md
```diff
@@ -23,9 +23,9 @@ But, everything has two sides, and so do digital protocols. The downsides are, t
 
 ## Supported Hardware
 
-DShot is supported on all BLHELI_S, BLEHLI_32 and KISS ESC’s. One limitation are older BLHELI_S ESC’s with EFM8BB1 MCU: Only DShot 150 and DShot 300 are supported on those, but this should still be good enough for 99% of use cases.
+DShot is supported by all BLHELI_S, BLHELI_32 and KISS ESCs. Note that older BLHELI_S ESC’s with EFM8BB1 MCUs only support DShot 150 and DShot 300, but this should be fine for most cases.
 
-No extra settings on the ESC’s are needed - they automatically detect by which protocol they are driven and act accordingly. Some firmware might only support a certain subset of protocols though, so be aware of that. Bluejay for example only supports DShot in all its variations, but none of the analog protocols.
+No additional configuration on the ESC-side is required to use DShot - they will automatically detect the protocol being used. Keep in mind that not all firmwares support all protocols, however. For example, Bluejay only supports DShot (all versions), but none of the analog protocols (e.g. PWM).
```
haslinghuis, line 28:
```suggestion
No additional configuration on the ESC-side is required to use DShot - they will automatically detect the protocol being used. Keep in mind that not all firmwares support all protocols, however. For example, Bluejay only supports DShot, but none of the analog protocols (e.g. PWM).
```

stylesuxx:
@ot0tot is correct with his assumption of the calculation being a typo - 53.28us is correct for DShot 300 frame length.

Regarding DShot 300 vs DShot 600. I don't know if I'd say *should be used* - since 300 is fine there too, and probably very few people will feel the difference anyway. I'd maybe re-phrase in such a way to make sure NOT to use DShot 600 with loop rates under 8k. Maybe encourage the user to test both and see if they notice a difference.

haslinghuis (approved):

nerdCopter (approved):

PR #399, opened 2024-03-27
Fixes markdown heading in CodingStyle.md

dellch:
## Issue
The "Braces are required" heading is not rendered on betaflight.com/docs/development/CodingStyle:
image

## This fix
This adds a space after `###` for the "Braces are required" heading, so it is rendered correctly on betaflight.com/docs/development/CodingStyle.  

It also adds an empty line after the heading, consistent with the other headings in the file.

## Further question
A few other locations below are not rendering code correctly - before I add that change: should I be creating issues first?  Any preference if it is in same PR or should be separate?

### DIFF
--- a/docs/development/CodingStyle.md
+++ b/docs/development/CodingStyle.md
@@ -89,7 +89,8 @@ if (x is true) {
 }
 ```
 
-###Braces are required
+### Braces are required
+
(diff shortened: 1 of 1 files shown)


dellch:
I mentioned some other inline code not being rendered in markdown, here is a screenshot:
image

haslinghuis (approved):

haslinghuis:
Thank you for contributing, please fix those in same PR here.

PR #446, opened 2024-07-26
Feature: Fancier board docs

VitroidFPV:
Adds an easy-to-write way to neatly show a product photo and important specs in a more visually important manner.

When adding a product image at the top of the doc previously, it would take up more space than the page even allows for unless it was included in a more complicated way. Now it integrates very well with the spec highlights and still follows the markdown image format.

Before:
image
After:
image


haslinghuis (approved):
Nice !!!

- How do we deal with boards with multiple hardware defines or board versions ?
- Noticed wire diagram seems upside down for Airbot Super F4

image

VitroidFPV:
> * How do we deal with boards with multiple hardware defines or board versions ?

There are multiple options we could take, in order of relative simplicity:
* Keep the main part of the document with the most current rev of the board and document all the changes made in the Notes tab
* Make the whole page tabbed with different board revs
* Make separate files for each rev which would somehow be linked together, though I'm not sure how at this point

> * Noticed wire diagram seems upside down for Airbot Super F4

I'll fix and update that as well, thanks for pointing it out

VitroidFPV:
Updated, should be ready to approve and merge

haslinghuis (approved):

blckmn (approved):

nerdCopter:
very nice.  i suspect the font-size could be smaller.
image

Issue #449, opened 2024-08-21
Changing EXTRA_PARAMS to EXTRA_FLAGS in How to Create a Target Configuration page


DanielsJusts:
Hello!

I found a mistake in the documentation. It references `EXTRA_PARAMS` but it is actually `EXTRA_FLAGS`.

https://github.com/betaflight/betaflight.com/blob/64e6231b695b2694521aedcd99d63581e10a6038/docs/development/manufacturer/creating-an-unified-target.md?plain=1#L45

PR #450, opened 2024-08-21
Update creating-an-unified-target.md

haslinghuis:
- fixes: #449


haslinghuis:
Reading again think we should update the paragraph for 4.5 which uses BOARD instead of TARGET

VitroidFPV (approved):

nerdCopter (approved):

PR #477, opened 2024-10-22
Update MCU requirements

haslinghuis:


ot0tot:
I agree this should be updated, and I have an opinion on what the default recommendations should be, but I hesitate to make them official without some testing/data to back them up. 

For example, I'm pretty sure the AT32F435 is ok at 8k, but would say never for the F411 at 8k. For F411/F405, the recommendation also depends on whether timer based DShot is being used. If the FC design requires bitbang or burst DMA to work (resulting in no gyro DMA), then a lower loop rate is almost certainly required.

docs/development/manufacturer/manufacturer-design-guidelines.md
```diff
@@ -460,20 +461,22 @@ Importantly, although the Bidirectional DShot ENABLED may require lower PID Loop
...
-|                               | BMI-270                               | 3.2 kHz       | Enabled                    | 3.2 kHz       | DShot 300      |
-| F411 SPI Rx \*\*\*            | MPU6000, MPU6050, ICM20601            | 8 kHz         | Enabled                    | 2 kHz         | DShot 300      |
-|                               | MPU6000, MPU6050, ICM20601            | 8 kHz         | Disabled (not recommended) | 4 kHz         | DShot 300      |
-|                               | BMI-270                               | 3.2 kHz       | Enabled                    | 1.6 kHz       | DShot 300      |
-|                               | BMI-270                               | 3.2 kHz       | Disabled (not recommended) | 3.2 kHz       | DShot 300      |
+| MCU                           | IMU                          | Sampling Rate | Bidirectional DShot Status | PID Loop Rate | DShot Protocol |
+| :---------------------------- | :--------------------------- | :------------ | :------------------------- | :------------ | :------------- |
+| H7XX                          | MPU60X0, ICM2060X, ICM42688P | 8 kHz         | Enabled or Disabled        | 8 kHz         | DShot 600      |
```
nerdCopter, line 466: while i understand this cleans up the tables, does it make searching harder?

haslinghuis: Other rows already used - so made it consistent.
But we need to decide on other factors like @ot0tot mentioned.

nerdCopter (approved):

blckmn (approved):

PR #471, opened 2024-10-05
Content: Add FLYSPARKF4

VitroidFPV:

PR #478, opened 2024-10-23
Add SKYSAKURAH743 board to the wiki

BloodSakura3774:

PR #480, opened 2024-10-28
Style: Fix info grid responsiveness, simplify info props

VitroidFPV:
As suggested in #471 - info grid shouldn't overflow across all screen widths, also allows for other blackbox storage media (flash, emulated SD card, SD card, external logger only) to be properly mentioned. 

Will move on to fix #471 and #478 after this - their preview builds will fail for the moment, but both will be converted as well

### DIFF
--- a/docs/development/manufacturer/fc_documentation/fc-doc-example.mdx
+++ b/docs/development/manufacturer/fc_documentation/fc-doc-example.mdx
@@ -7,7 +7,7 @@ sidebar_custom_props:
     imu: ICM42688P
     osd: MAX7456
     barometer: DPS310
-    flash: W25Q128JVPQ (16MB)
+    blackbox: Onboard 16MB
     dimensions: 36x39mm
     mounting: 30.5x30.5mm
     weight: 7.8g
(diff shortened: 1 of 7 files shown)


docs/development/manufacturer/fc_documentation/fc-doc-example.mdx
```diff
@@ -7,7 +7,7 @@ sidebar_custom_props:
     imu: ICM42688P
     osd: MAX7456
     barometer: DPS310
-    flash: W25Q128JVPQ (16MB)
+    blackbox: Onboard 16MB
```
haslinghuis, line 10: IMO `Onboard` is redundant here.

haslinghuis (dismissed):

nerdCopter:
data font is still dynamically sized. i feel it needs to be static/smaller.
and is also wrapping depending on window-width image
that said, i think it's better than before.

docs/development/manufacturer/fc_documentation/fc-doc-example.mdx
haslinghuis: Agree with @nerdCopter font can be a little more smaller and content static.

VitroidFPV:
It appears to be largely based on the system font as that's what Docusaurus uses as a basis for everything. On my end the system font does not overflow, so I wasn't able to catch this. I will reduce it further to catch various fonts

nerdCopter (approved):

VitroidFPV:
Sorry, forgot to include both changes in the commit 😅

haslinghuis (approved):

nerdCopter (approved):

PR #485, opened 2024-11-06
Add warning for led pin resources

haslinghuis:


VitroidFPV:
The limit on PC13-15 is mostly consistent across all STM32 and AT32 MCUs, the wording from the datasheet is well written, makes sense to use it here. LGTM

VitroidFPV (approved):

PR #492, opened 2024-12-02
Clarify Low Power Disarm option

Agilulfo:
- Document how the lowest power lever is picked


docs/wiki/configurator/vtx-tab.md
```diff
@@ -25,7 +25,7 @@ It is important to have VTX Telemetry (SmartAudio or Tramp) enabled in the Ports
...
-| **Low Power Disarm**         | When enabled, the VTX uses the lowest available power when disarmed (except if a failsafe has occurred).<li>**On** - Go to lowest power possible when disarmed</li><li>**Off** - Always on the set power level</li><li>**On until first arm** - Stay on the lowest power possible until the first time the quad is armed, then go to the set power and stay there indefinitely</li> |
+| **Low Power Disarm**         | When enabled, the VTX uses the lowest available power when disarmed (except if a failsafe has occurred).<li>**On** - Go to lowest power possible when disarmed</li><li>**Off** - Always on the set power level</li><li>**On until first arm** - Stay on the lowest power possible until the first time the quad is armed, then go to the set power and stay there indefinitely</li> Note: The lowest power level is assumed to be the first defined in the power levels list.|
```
Agilulfo, line 28: This seems to me to be the current behavior. Would love somebody to confirm.
It was very confusing for me because in my quad for some reason the manufacturer configured the PIT mode as last in the power levels list.
so to get the quad start in PIT level I had to reorder the levels.

Agilulfo:
Please let me know if it is first necessary to create an issue.

docs/wiki/configurator/vtx-tab.md
```diff
@@ -25,7 +25,7 @@ It is important to have VTX Telemetry (SmartAudio or Tramp) enabled in the Ports
...
 | **Pit Mode**                 | When enabled, the VTX enters in a very low power mode to let the quad be on at the bench without disturbing other pilots. Usually the range of this mode is less than 5m.<br/>NOTE: Some protocols, like SmartAudio, can't enable Pit Mode via software after power-up.                                                                                                             |
 | **Pit Mode frequency**       | Select the frequency to use in Pit Mode.                                                                                                                                                                                                                                                                                                                                            |
```
Agilulfo, line 27: Maybe it sounds obvious but thinking to add a "Note" here too to say that 0 means to stay in the current frequency

docs/wiki/configurator/vtx-tab.md
```diff
@@ -25,7 +25,7 @@ It is important to have VTX Telemetry (SmartAudio or Tramp) enabled in the Ports
 | **Power**                    | Select the power level to use.                                                                                                                                                                                                                                                                                                                                                      |
 | **Pit Mode**                 | When enabled, the VTX enters in a very low power mode to let the quad be on at the bench without disturbing other pilots. Usually the range of this mode is less than 5m.<br/>NOTE: Some protocols, like SmartAudio, can't enable Pit Mode via software after power-up.                                                                                                             |
```
Agilulfo, line 26: it is unclear here what are the implications of the Note in here.

- does it mean I cannot enable pit mode from the configurator if I use SmartAudio?
- how do I enable PitMode to be on when I power my craft?

nerdCopter: - correct, SmartAudio cannot enable pitmode via configurator.
- i suppose it depends on what hardware you have.  maybe smart-audio does not have true pirmode, but rather only a lowest setting -- i am unsure and defer to some racers to answer.  @limonspb ?

docs/wiki/configurator/vtx-tab.md
nerdCopter: seems logical.
additionally, manufacturers often do things incorrectly.

Agilulfo: I'm also experiencing an issue with this feature that works only partially. 
If I use the "on until first arm" option the quad start with the lowest power  as expected but it does not change automatically for the predefined value. 
Once disarmed I can manually change the power (e.g. via switch) but the change does not happen automatically.

I remember I've read somewhere that this feature works only with some VTXs but I have not yet found evidence of this in the issue list of this repo.
I'll open a bug report but if this turns out to be a known issue then I think it is worth to document it here.

docs/wiki/configurator/vtx-tab.md
nerdCopter: i'm unsure that this is correct.  looking at some old code (BF 3.5, 4.0), pitmode 0 defined as setting "out-of-range"; however, maybe this is true for modern VTX tables. i am unsure.

docs/wiki/configurator/vtx-tab.md
nerdCopter: i do not think low_power_disarm should be used when using vtx power on a switch. the options should probably be used mutually exclusive.

Agilulfo: I've just tested it.
I've removed the vtx commands that allowed me to change the power with my switch but low_power_disarm set to "on until first arm" is still staying at the lowest power.

nerdCopter: what hardware, i beleive there are several reports of problems with HDZ AIO5 by CrazyBee

Agilulfo: it is the AIO (4in1) from [betafpv](https://betafpv.com/products/air-brushless-flight-controller?variant=41142912712838)  for the air65

Agilulfo:
This PR is getting stale. 

I've got the feeling that the proposed changes are not getting traction and perhaps the clarifications I would like to add are not relevant. 

I'll give it a couple of more days and I'll close the PR. Feel free to close earlier than that if prefer.

limonspb:
@SupaflyFPV could you plz check this, as a main BF wiki boss :)

PR #558, opened 2025-07-25
Using of curves MinMax values in workspace

demvlad:
Added manual how to use curves MinMax values in workspace


haslinghuis (commented):
Thanks.

Images should go in static/img/blackbox.

haslinghuis (approved):

PR #573, opened 2025-09-30
Correct bytes back to bits for the flash USE flag description

osirisinferi:
I noticed that in #523 a few of the kilo*bit*/mega*bit* abbreviations were incorrectly corrected to kilo*byte* and mega*byte*.

This PR corrects the abbreviations (kilobit was never correct by the way, as the W25M512 chip has a 512 *mega*bit size), adds the *byte* value behind the *bit* value to prevent confusion and also explains that the `USE_FLASH_W25Q128FV` flag is only necessary for quad and/or octo SPI. (Adding `USE_FLASH_W25Q128FV` without actually using quad/octo SPI only increases the FW unnecessarily). Also, no `W25Q8` chip is part of the W25Q128FV driver, however `W25Q16JV` is.

By the way, all that non-flash USE flag changes were done by the `npx lint-staged` command which automatically ran when I commited those flash USE flag changes... Not sure if you'd all want that; can try to remove those changes if necessary.

### DIFF
--- a/docs/development/manufacturer/manufacturer-design-guidelines.md
+++ b/docs/development/manufacturer/manufacturer-design-guidelines.md
@@ -16,27 +16,27 @@ import useBaseUrl from '@docusaurus/useBaseUrl'
 
 ## Version Change Register
 
-| Version # | Revision Date    | Changes, Reasons, and Notes       |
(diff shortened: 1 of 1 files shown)


nerdCopter (approved):
prefer refactoring separately, but good to go.

PR #576, opened 2025-10-07
UPDATE: Renaming configurator to app for consistency

blckmn:


haslinghuis (approved):

nerdCopter (approved):
but we will always call it configurator 😹

PR #694, opened 2026-05-22
fix(magnetometer): default address `14` instead of `12`

jmevel:
Default address had been changed in betaflight/betaflight#13995

I also removed the mention of `the manufacturer's recommended default` address. I went through the IST8310 datasheet and couldn't find any mention of it.

## Summary by CodeRabbit

* **Documentation**
  * Updated Magnetometer guide with corrected IST8310 default i2c address and fallback address values for configuration troubleshooting.


haslinghuis (dismissed):

haslinghuis (approved):

jmevel:
@haslinghuis seems this PR is OK to merged. Would that be possible to do it? 
Thanks

PR #718, opened 2026-07-10
Added ADRC documentation.

bvandevliet:
See https://github.com/betaflight/betaflight/pull/15400

Although the feature is not yet released at all, I decided to mark this PR as "ready for review". Hoping to gain more exposure of this feature being developed and reaching more pilots trying and testing it, when docs are already available upfront.

## Summary by CodeRabbit

* **Documentation**
  * Added a guide to the experimental ADRC flight-control mode as an opt-in alternative to classic PID rate control.
  * Documented CLI configuration, tunables, defaults, and valid ranges.
  * Included tuning guidance, debug logging, operational constraints, hardware considerations, testing cautions, and known limitations.
  * Described runtime safeguards, state resets, and behavior during flight-state transitions.


docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
+# ADRC (Active Disturbance Rejection Control)
+
+ADRC is an experimental, opt-in alternative to classic PID rate control, selected per PID profile via `pid_type`. Instead of proportional/integral/derivative gains acting on error, ADRC runs a second-order **Extended State Observer (ESO)** that continuously estimates the craft's rotation rate, its derivative, and a lumped "everything else" disturbance term (motor/prop mismatch, wind, payload imbalance, CG offset), then drives a virtual PD law to cancel it. In principle this rejects disturbances faster and needs less per-craft tuning than classic PID.
```
danusha2345: Terminology nit: the observer itself is **third**-order (three states: z1/z2/z3) — "second-order" describes the ADRC/plant model it observes, not the ESO. The three-state list right below this sentence would otherwise contradict it.

```suggestion
ADRC is an experimental, opt-in alternative to classic PID rate control, selected per PID profile via `pid_type`. Instead of proportional/integral/derivative gains acting on error, ADRC runs a third-order **Extended State Observer (ESO)** that continuously estimates the craft's rotation rate, its derivative, and a lumped "everything else" disturbance term (motor/prop mismatch, wind, payload imbalance, CG offset), then drives a virtual PD law to cancel it. In principle this rejects disturbances faster and needs less per-craft tuning than classic PID.
```

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
...
+
+:::info
+**Not yet in an official Betaflight release.** This page documents an open pull request, [betaflight/betaflight#15400](https://github.com/betaflight/betaflight/pull/15400) — `pid_type = ADRC` does not exist in any stock Betaflight build (Configurator releases, official `master`) until that PR merges; `set pid_type = ADRC` will simply error out as an unknown value on a normal build. To try it now, flash one of the PR's [prebuilt hex releases](https://github.com/danusha2345/ADRC-betaflight/releases).
```
danusha2345: Micro-nit: on a stock build the `pid_type` variable doesn't exist at all, so the CLI error is "unknown name", not "unknown value". Worth being exact so users recognize the actual error message they'll see.

```suggestion
**Not yet in an official Betaflight release.** This page documents an open pull request, [betaflight/betaflight#15400](https://github.com/betaflight/betaflight/pull/15400) — `pid_type = ADRC` does not exist in any stock Betaflight build (Configurator releases, official `master`) until that PR merges; `set pid_type = ADRC` will simply error out on a normal build — the CLI rejects `pid_type` as an unknown setting name, since the variable itself doesn't exist there. To try it now, flash one of the PR's [prebuilt hex releases](https://github.com/danusha2345/ADRC-betaflight/releases).
```

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
...
... ESO's `b0·u` feedback is held at zero until liftoff is detected (`adrc_liftoff_throttle`, default 40%; or sustained rotation above `adrc_liftoff_gyro_dps`, default 20°/s, held for `adrc_liftoff_hold_ms`, default 25 ms — a toss launch opens it almost instantly). Without this, the observer winds up while grounded and has to unwind violently at liftoff. Once open, **the gate stays open until disarm — there is no mid-air re-arm.** A repeated ground test (e.g. bench reps) needs an actual disarm/re-arm cycle, not just a return to idle throttle: an earlier opt-in re-arm heuristic was removed rather than kept, because throttle+gyro alone can't reliably distinguish a real landing from a calm mid-air float — see `ADRC-020` in the [remediation tracker](https://github.com/danusha2345/ADRC-betaflight/blob/master/docs/ADRC_REMEDIATION_TRACKER.md) for the full reasoning. **`adrc_liftoff_throttle` has no built-in relationship to `adrc_hover_throttle`** below — they answer different questions ("how sure am I this throttle means I'm off the ground" vs. "where do I actually hover"). Set `adrc_liftoff_throttle` a bit _above_ your actual hover throttle rather than assuming the default fits your craft.
```
danusha2345: Sequencing: as of the current PR head (`79f8b60`) the `adrc_liftoff_idle_*` params and the re-arm branch still exist in the code — the removal is decided (`ADRC-020`) but deliberately lands only after the b4 verification flight, to keep the byte-identical A/B property. Suggest present-progressive here so the page can't contradict the code if it merges first (the CLI table below already omits the idle params, which is right for the target state).

```suggestion
- **Liftoff gate** — while the craft is ground-constrained, the plant doesn't respond to output the way the model expects, so the ESO's `b0·u` feedback is held at zero until liftoff is detected (`adrc_liftoff_throttle`, default 40%; or sustained rotation above `adrc_liftoff_gyro_dps`, default 20°/s, held for `adrc_liftoff_hold_ms`, default 25 ms — a toss launch opens it almost instantly). Without this, the observer winds up while grounded and has to unwind violently at liftoff. Once open, **the gate stays open until disarm — there is no mid-air re-arm.** A repeated ground test (e.g. bench reps) needs an actual disarm/re-arm cycle, not just a return to idle throttle: an earlier opt-in re-arm heuristic is being removed rather than kept (decided; the code change lands after the in-progress verification flight), because throttle+gyro alone can't reliably distinguish a real landing from a calm mid-air float — see `ADRC-020` in the [remediation tracker](https://github.com/danusha2345/ADRC-betaflight/blob/master/docs/ADRC_REMEDIATION_TRACKER.md) for the full reasoning. **`adrc_liftoff_throttle` has no built-in relationship to `adrc_hover_throttle`** below — they answer different questions ("how sure am I this throttle means I'm off the ground" vs. "where do I actually hover"). Set `adrc_liftoff_throttle` a bit _above_ your actual hover throttle rather than assuming the default fits your craft.
```

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
...
+- **Gated z3 decay** (`adrc_gated_z3_decay`, default 20/s) — even with the gate above, z3 is still a leaky integrator of the observer error regardless of gate state; while grounded it decays at this rate (always faster than the airborne `adrc_sigma_decay`) so it can't quietly wind up during a long armed-idle period even without exceeding the gate's rotation threshold.
```
danusha2345: Precision nit: the code takes `max(gated, sigma_decay, 1.0)` — so it's "never slower" (equal is possible), with a 1/s floor even if both are set to 0.

```suggestion
- **Gated z3 decay** (`adrc_gated_z3_decay`, default 20/s) — even with the gate above, z3 is still a leaky integrator of the observer error regardless of gate state; while grounded it decays at this rate (never slower than the airborne `adrc_sigma_decay`, and floored at 1/s) so it can't quietly wind up during a long armed-idle period even without exceeding the gate's rotation threshold.
```

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
...
... tunables yet — you read and set all of them (`adrc_wc`/`adrc_wo`/`adrc_b0`, the gyro filter, the liftoff-gate thresholds, everything in the [CLI Reference](#cli-reference) below) through the CLI tab, `set`/`get`, or `diff`/`dump`. Meanwhile the Configurator's **PID Tuning tab** (P/I/D/F sliders, D Max, TPA) and the D-term-specific rows on the **Filter tab** (`dterm_lpf1`/`lpf2`) don't grey out, don't hide, and don't warn you — they keep computing and displaying numbers exactly as if you were still on classic PID. None of that computation reaches the mixer: `pid.c` runs the classic P/I/D/D-term-filter/TPA calculation unconditionally regardless of `pid_type` (to keep classic PID's code path byte-identical), then simply overwrites the result with ADRC's output before it's used, on an ADRC profile. Anti-gravity is the one exception — it's explicitly skipped rather than computed-then-discarded — but the practical effect is the same: **drag any of those sliders on an ADRC profile and the craft won't respond, with nothing in the UI telling you why.** (The throttle-scaled `b0` — `adrc_hover_throttle`/`adrc_b0_scale_max` — is ADRC's own equivalent of TPA, set via CLI like everything else.)
```
danusha2345: There are two explicit skips in `pid.c`, not one: anti-gravity and the zero-throttle I-term reset handling (`pid.c` guards it with `pid_type != ADRC`; ADRC has its own gated-decay mechanism instead).

```suggestion
**ADRC is CLI-only, and every classic-PID Configurator control turns into a zombie on an ADRC profile.** There's no dedicated GUI screen for `adrc_*` tunables yet — you read and set all of them (`adrc_wc`/`adrc_wo`/`adrc_b0`, the gyro filter, the liftoff-gate thresholds, everything in the [CLI Reference](#cli-reference) below) through the CLI tab, `set`/`get`, or `diff`/`dump`. Meanwhile the Configurator's **PID Tuning tab** (P/I/D/F sliders, D Max, TPA) and the D-term-specific rows on the **Filter tab** (`dterm_lpf1`/`lpf2`) don't grey out, don't hide, and don't warn you — they keep computing and displaying numbers exactly as if you were still on classic PID. None of that computation reaches the mixer: `pid.c` runs the classic P/I/D/D-term-filter/TPA calculation unconditionally regardless of `pid_type` (to keep classic PID's code path byte-identical), then simply overwrites the result with ADRC's output before it's used, on an ADRC profile. Anti-gravity and the zero-throttle I-term reset heuristic are the exceptions — they're explicitly skipped rather than computed-then-discarded — but the practical effect is the same: **drag any of those sliders on an ADRC profile and the craft won't respond, with nothing in the UI telling you why.** (The throttle-scaled `b0` — `adrc_hover_throttle`/`adrc_b0_scale_max` — is ADRC's own equivalent of TPA, set via CLI like everything else.)
```

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
...
+
+**The rest of the Filter tab is not a zombie — it's fully live for ADRC too.** The dynamic notch, RPM filter, and `gyro_lpf1`/`lpf2` static/dynamic lowpass all run in the gyro-sampling task, upstream of and completely independent from the PID loop (`gyro.c` has no knowledge of `pid_type` at all), so they affect ADRC's ESO input exactly as they affect classic PID's. Only the D-term-specific rows above are dead; everything else on that tab still matters and should be tuned as normal.
+
+**Simplified Tuning is the one control that's actually harmless, not just inert-and-silent, and it's worth knowing why.** danusha2345's fork repurposes the classic `p`/`i`/`d` sliders as its ADRC gains, so on that fork Simplified Tuning (which recalculates those same fields) has to be manually disabled or it silently overwrites the tune on save. This implementation uses dedicated `adrc_wc`/`adrc_wo`/`adrc_b0` fields instead specifically to avoid that trap: Simplified Tuning only ever recalculates `pid[axis].P/I/D/F` and `d_max[axis]` — the same already-zombie fields described above — so on an ADRC profile it's just recomputing numbers nobody reads, never the `adrc_*` fields themselves.
```
danusha2345: This is the one substantive correction: `simplified_tuning.c` recalculates more than `pid[axis].P/I/D/F` + `d_max`. Its D-term filter multiplier rewrites the `dterm_lpf1`/`lpf2` cutoffs (zombies under ADRC, harmless), but its **gyro filter multiplier rewrites the `gyro_lpf1` cutoffs — which are fully live for ADRC** per the Filter-tab paragraph above (they shape what the ESO sees). So "harmless" holds for the PID sliders, not for Simplified Tuning as a whole. Suggested heading fix below; the following sentence's "only ever recalculates `pid[axis].P/I/D/F` and `d_max[axis]`" would also need scoping to the PID sliders, plus one closing sentence, e.g.: *"The filter sliders are a different story: the D-term multiplier rewrites already-zombie `dterm_lpf` cutoffs, but the gyro multiplier rewrites `gyro_lpf1` — live for ADRC like the rest of the gyro chain — so that one slider still changes what the ESO sees."*

```suggestion
**Simplified Tuning's PID sliders are actually harmless, not just inert-and-silent, and it's worth knowing why — but its gyro-filter slider is live.** danusha2345's fork repurposes the classic `p`/`i`/`d` sliders as its ADRC gains, so on that fork Simplified Tuning (which recalculates those same fields) has to be manually disabled or it silently overwrites the tune on save. This implementation uses dedicated `adrc_wc`/`adrc_wo`/`adrc_b0` fields instead specifically to avoid that trap: Simplified Tuning only ever recalculates `pid[axis].P/I/D/F` and `d_max[axis]` — the same already-zombie fields described above — so on an ADRC profile it's just recomputing numbers nobody reads, never the `adrc_*` fields themselves.
```

danusha2345 (commented):
Reviewed against the actual PR head (`betaflight/betaflight#15400` @ `79f8b6041d`), claim by claim. Verified as accurate: the full CLI table (defaults and ranges incl. the non-obvious `adrc_wo_yaw = 80`), the debug channel map (incl. z3 ÷16 and the sign-tagged gate in [7]), the blackbox headers (`pid_type` + every `adrc_*` field), the computed-then-overwritten classic path and the zombie/live Configurator split, the `wo·dT ≤ 0.5` hardening, F446/F411 notes, the thrust-linear behavior, the whoop data point (33/65/3200 — correctly converted from the fork's 33/65/160 @ scale 20), and every cross-link/anchor.

Seven inline suggestions below — five are wording/precision-level; two are worth attention:

1. **ADRC-020 tense** (liftoff-gate bullet): the page describes the opt-in re-arm as already removed, but at the current PR head the `adrc_liftoff_idle_*` params still exist — the removal is decided and lands after the verification flight. Suggested present-progressive wording so the page and the code can't contradict each other regardless of merge order.
2. **Simplified Tuning** is not harmless as a whole: its gyro-filter multiplier rewrites `gyro_lpf1` cutoffs, which are live for ADRC (they shape the ESO input). The PID-slider part of the claim is correct.

Great page overall — happy to see the tracker/fork wired in as the live-status home so this stays evergreen.

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
...
+
+| Classic PID concept                                                        | Closest ADRC field                      | Raising it                                                            | Too high                                               | Too low                               |
+| -------------------------------------------------------------------------- | --------------------------------------- | --------------------------------------------------------------------- | ------------------------------------------------------ | ------------------------------------- |
+| P and D, raised together (locked ratio, can't trade one against the other) | `adrc_wc` — controller bandwidth        | Crisper, faster correction, shorter settling time                     | Oscillation/overshoot, motors "singing" on throttle-up | Floaty, sluggish                      |
+| I (continuous, non-integrating) fused with the D-term filter cutoff        | `adrc_wo` — observer bandwidth          | Faster rejection of wind/prop-wash/CG-offset disturbance              | Gyro noise amplification — hover chatter/heat          | Laggy, "soft" recovery from bumps     |
```
danusha2345: Wording: z3 *is* an integrator — it accumulates the observer innovation (`z3' = -β3·e`), leaky only when `adrc_sigma_decay > 0`. "Non-integrating" would tell a classic-PID pilot the exact opposite of the windup-clamp discussion in the Numerical-hardening bullet above.

```suggestion
| I (continuously estimated by the observer, optionally self-decaying) fused with the D-term filter cutoff        | `adrc_wo` — observer bandwidth          | Faster rejection of wind/prop-wash/CG-offset disturbance              | Gyro noise amplification — hover chatter/heat          | Laggy, "soft" recovery from bumps     |
```

docs/wiki/guides/current/ADRC.md
bvandevliet: I'm okay having this small window of incorrectness, it prevents us to need to update the docs after b5 is build.

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
...
+
+One structural consequence worth internalizing: because `wc` locks P:D together and `wo` fuses I-and-filter into one number, ADRC has one fewer independently-tunable "feel" degree of freedom than classic 3-term PID, for a given axis. That's an intentional simplification (Gao's standard bandwidth-parameterized ADRC), not an oversight — the tradeoff for fewer, less-interacting knobs.
+
+### Tuning procedure
+
+A sensible starting order, adapted from the community procedure on danusha2345's fork:
+
+1. **`adrc_b0`** — raise until the craft takes off stably and responds crisply; keep raising until you hear stuttering/chatter in hover, then back off ~20%. Over-estimating is fairly harmless; under-estimating causes instability.
```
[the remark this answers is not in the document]
bvandevliet: Line 118 finding (adrc_b0 direction): skipped, not valid. Verified against @danusha2345's actual README (the explicit source for this procedure) — it says "raise until stable, keep raising until stuttering, back off ~20%; overestimating is harmless, underestimating causes instability," word-for-word matching the current doc. Flipping to "lower toward instability" would invert the real procedure and contradict the b0 semantics stated elsewhere in the same file.

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
+# ADRC (Active Disturbance Rejection Control)
+
+ADRC is an experimental, opt-in alternative to classic PID rate control, selected per PID profile via `pid_type`. Instead of proportional/integral/derivative gains acting on error, ADRC runs a third-order **Extended State Observer (ESO)** that continuously estimates the craft's rotation rate, its derivative, and a lumped "everything else" disturbance term (motor/prop mismatch, wind, payload imbalance, CG offset), then drives a virtual PD law to cancel it. In principle this rejects disturbances faster and needs less per-craft tuning than classic PID.
```
danusha2345, line 3: Nit, and it's the code that's wrong rather than the page: "third-order Extended State Observer" here vs `adrc.h`'s header comment "A second-order linear ESO". Three states → third-order observer on a second-order plant, so this page uses the standard LADRC wording; worth fixing the `adrc.h` comment so the two don't contradict each other.

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
...
+
+:::info
+**Not yet in an official Betaflight release.** This page documents an open pull request, [betaflight/betaflight#15400](https://github.com/betaflight/betaflight/pull/15400) — `pid_type = ADRC` does not exist in any stock Betaflight build (Configurator releases, official `master`) until that PR merges; `set pid_type = ADRC` will simply error out on a normal build — the CLI rejects `pid_type` as an unknown setting name, since the variable itself doesn't exist there. To try it now, flash one of the PR's [prebuilt hex releases](https://github.com/danusha2345/ADRC-betaflight/releases).
```
danusha2345: The link lands on the release list, whose newest entry is now `adrc-pr15400-b5` (2026-07-18) — and that build is PR head `eda3bb16eb` **plus** a fork-side per-profile `adrc_b0_law` selector (QUADRATIC/SQRT/LINEAR/FIXED) which is neither in this branch nor in the CLI table below. A tester who flashes it meets an undocumented setting and has no way to know which law to report.

Also worth one sentence: b5 is **PG 0**, not 16 — the PG version is a 4-bit field (`pg.h`, `PGR_PGN_VERSION_MASK = 0xf000`, `_version << 12`) and 15 was already the maximum, so it wraps. Practical consequence for a tester: flashing b5 over a b2–b4 install *or* over a PR-head build **resets the PID profiles to defaults** — `diff all` first.

Simplest fix is to name the tag you want testers on (b4 matches this page's tunable set exactly) and add a line noting b5's extra selector + the profile reset.

danusha2345 (commented):
Reviewed this page line by line against the PR branch as it stands (`bvandevliet:adrc-toggle` @ `eda3bb16eb`) rather than against my own docs, since the page is written from the code. It holds up remarkably well — the parts I checked and found exact:

* **The whole CLI table.** All 18 settings from `settings.c:1417-1434` are present, and every default and range matches `adrcResetProfile()` (`adrc.c:148-231`). Nothing missing, no wrong number.
* **Debug channels [0]–[7]**, including the sign-tagged gate in [7] — matches `adrc.c:519-530`.
* **The control law** `u = (kp·(sp − z1) − kd·z2 − z3)/b0`, `kp = wc²`, `kd = 2wc`, P taken from the TD-filtered setpoint — matches `adrc.c:501-505`.
* **Numerical hardening**: `wo·dT ≤ 0.5`, z1/z2 as anti-divergence bounds only, z3 as the one authority-clamped state, the `-ffast-math`-proof finite check.
* **The Configurator "zombie" section**, including that classic P/I/D is computed unconditionally and overwritten, and that anti-gravity and the zero-throttle I-term reset are the two explicit exceptions (`pid.c:1512-1524`, `1577`, `1119`). This is the part most likely to save a tester's evening and it's accurate.
* **F446 exclusion** and the "persisted layout unaffected" claim; **`pid_type` genuinely doesn't exist in stock** (grepped current upstream `master`, zero hits); **blackbox headers** do log `pid_type` plus every `adrc_*` field.
* **Everything it cites from my fork resolves**: README, `ADRC_FIXES.md`, the tracker and its three anchors, both analysis scripts. The whoop conversion is right too (`160 × adrc_b0_scale 20 = 3200`), and the shipped defaults really do trace to the round-2 sweep (`ADRC_FIXES.md`, "the 5\" sweet spot was 60/100/200").

Four small factual corrections and one safety gap are in the inline comments. The only one I'd call blocking is **ADRC-026**, which postdates this page and is directly in tension with tuning step 2 ("raise `wo` until chatter appears in hover") — that path can open the liftoff gate while the craft is still on the ground at zero stick throttle.

Housekeeping on my side, since reviewing your page surfaced it: the fork's example-tune table listed the older wc=30 row as "blackbox-refined" while the round-2 sweep this page (correctly) cites converged on wc=60; ADRC-020 was still filed under "Open items" despite being CLOSED; the PG lineage note stopped at "current head = PG 15" and predicted a b5 inheriting it (b5 is actually PG 0 — the version field is 4 bits, `pg.h`'s `PGR_PGN_VERSION_MASK 0xf000`, so it wrapped); and How-to-help-test carried no ADRC-026 ground warning. All four fixed in danusha2345/ADRC-betaflight@453225a065, so the links from this page stay consistent.

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
...
+- **Gated z3 decay** (`adrc_gated_z3_decay`, default 20/s) — even with the gate above, z3 is still a leaky integrator of the observer error regardless of gate state; while grounded it decays at this rate (never slower than the airborne `adrc_sigma_decay` — a low or zero `adrc_gated_z3_decay` is floored up to match it, rather than actually applied) so it can't quietly wind up during a long armed-idle period even without exceeding the gate's rotation threshold.
```
danusha2345: There's an absolute floor as well, not just "never slower than `adrc_sigma_decay`": `adrc.c:262` is `fmaxf(fmaxf(gatedZ3Decay, c->decayRate), 1.0f)`. So with `adrc_gated_z3_decay = 0` and the default `adrc_sigma_decay = 3` (0.3/s) the grounded decay is **1.0/s**, not the airborne rate — the code comment puts it as "Ground tau is floored at ~1 s". Same wording appears in the CLI table row (line 89).

While here: `adrc_sigma_decay = 3` is 0.3/s, i.e. τ ≈ 3 s — quoting the time constant alongside the ×0.1 rate would make the row easier to reason about.

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
...
+
+**Simplified Tuning's PID sliders are actually harmless, not just inert-and-silent, and it's worth knowing why — but its gyro-filter slider is live.** @danusha2345's fork repurposes the classic `p`/`i`/`d` sliders as its ADRC gains, so on that fork Simplified Tuning (which recalculates those same fields) has to be manually disabled or it silently overwrites the tune on save. This implementation uses dedicated `adrc_wc`/`adrc_wo`/`adrc_b0` fields instead specifically to avoid that trap: Simplified Tuning only ever recalculates `pid[axis].P/I/D/F` and `d_max[axis]` — the same already-zombie fields described above — so on an ADRC profile it's just recomputing numbers nobody reads, never the `adrc_*` fields themselves.
+
+**`pid_type` is a disarmed-only configuration change**, exactly like editing any other field in a PID profile — it is not a supported in-flight handover between control laws. Switching a profile's `pid_type` (or switching to a different profile with a different `pid_type`) clears both the classic I-term and the full ADRC state (observer + gate) on the transition, so don't rely on toggling into or out of ADRC while armed.
```
danusha2345: Precise version: on a `pid_type` change `pid_init.c` clears classic I unconditionally but resets the gate only under `if (!ARMING_FLAG(ARMED))` — an armed transition deliberately leaves the liftoff latch alone (the comment there explains why: adjustment ranges legitimately re-run `pidInitConfig()` in flight). The paragraph's advice is right; it's just "(observer + gate)" that doesn't hold for the armed case.

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
...
+
+One structural consequence worth internalizing: because `wc` locks P:D together and `wo` fuses I-and-filter into one number, ADRC has one fewer independently-tunable "feel" degree of freedom than classic 3-term PID, for a given axis. That's an intentional simplification (Gao's standard bandwidth-parameterized ADRC), not an oversight — the tradeoff for fewer, less-interacting knobs.
+
+### Tuning procedure
+
+A sensible starting order, adapted from [@jmsweng's community tuning procedure](https://github.com/danusha2345/ADRC-betaflight/blob/master/README.md#tuning-procedure-community-from-jmsweng) on @danusha2345's fork:
+
+1. **`adrc_b0`** — raise until the craft takes off stably and responds crisply; keep raising until you hear stuttering/chatter in hover, then back off ~20%. Over-estimating is fairly harmless; under-estimating causes instability.
+2. **`adrc_wo`** — raise until chatter appears in hover (the observer starting to track gyro noise), then back off ~20%.
```
danusha2345: This step now collides with a finding that postdates the page — **ADRC-026**, from the b5 wc/wo 2×2 flights: at `wo = 150` the craft self-oscillates at ~28.5 Hz on the ground at idle, which the gyro-only liftoff path (`adrc_liftoff_gyro_dps = 20` sustained for `adrc_liftoff_hold_ms = 25`) reads as liftoff within 0.1–0.4 s of arming. The gate then opens **with the craft on the ground at 0 % stick throttle**, z3 winds up on ground-contact dynamics it can't model, and the motors run to saturation — 3 of 5 arms at that `wo` show the signature; the other two were disarmed before the detector fired.

"Raise `wo` until chatter appears in hover" is exactly the path into it, so a caveat here (and a Testing Notes bullet — see next comment) would be worth having before the page goes live. Written up under ADRC-026 in [the tracker](https://github.com/danusha2345/ADRC-betaflight/blob/master/docs/ADRC_REMEDIATION_TRACKER.md), data in `docs/flight-test-analysis/pr15400-b5-wcwo2x2/`.

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,173 @@
...
+- **`pid_at_min_throttle = OFF`** disables the controller entirely below the arming throttle threshold — sticks/tilts produce zero motor response until throttle is raised. The default `ON` is safe under ADRC because the liftoff gate already provides ground protection; only turn it off if you specifically want the "fully asleep at idle" behavior and understand the trade-off.
+- **`dshot_bidir = ON`** has been observed to cause hard gyro freezes (thousands of identical gyro samples, momentarily uncontrollable) on at least one board/firmware-base combination. If your craft flies erratically or "acts possessed," try `set dshot_bidir = OFF` (you lose the RPM filter; the dynamic notch still works).
+- If you want to compare efficiency (current draw) against classic PID, blackbox logs `energyCumulative` per flight — fly matched maneuvers on both `pid_type`s and compare. The props-off heat finding above doesn't generalize to real flight either way.
```
danusha2345, line 157: Suggested extra bullet in the same register as the `dshot_bidir` one above:

> **If the motors are audibly oscillating at idle right after arming, disarm immediately.** At high `adrc_wo` a self-induced ground oscillation can trip the liftoff detector's gyro path, after which the observer spools the motors up with your throttle still down (ADRC-026). Arm props-off first when trying a higher `wo`.

I've just added the same warning to the tracker's How-to-help-test section, so the two funnels stay consistent.

bvandevliet:
@danusha2345 Thanks for re-reviewing! I've addressed all you comments, and corrected the `adrc.h` header comment in the firmware branch too. I did however push a follow-up commit though because I noticed by addressing your comments I re-introduced tight coupling again with external state of ADRC development. Should be all good now.

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,174 @@
...
+- **Gated z3 decay** (`adrc_gated_z3_decay`, default 20/s) — even with the gate above, z3 is still a leaky integrator of the observer error regardless of gate state; while grounded it decays at this rate (never slower than the airborne `adrc_sigma_decay` — a low or zero `adrc_gated_z3_decay` is floored up to match it, rather than actually applied) so it can't quietly wind up during a long armed-idle period even without exceeding the gate's rotation threshold. There's also an **absolute floor of 1.0/s** on top of that comparison (`adrc.c`: `fmaxf(fmaxf(gatedZ3Decay, decayRate), 1.0f)`) — with the defaults (`adrc_gated_z3_decay` low/zero, `adrc_sigma_decay = 3` → 0.3/s airborne), grounded decay is floored at **1.0/s** (τ ≈ 1 s), not at the slower airborne rate.
```
danusha2345: Two problems here, and the second one comes from my earlier review comment rather than from your edit.

1. "with the defaults (`adrc_gated_z3_decay` low/zero, ...)" — the default is `200` → 20/s, so the 1.0/s floor never binds at defaults; it only binds once a tester lowers the setting.
2. More importantly, both this bullet and my original comment say the fast decay applies "while grounded". The code switches on the **latch**, not on physical ground contact: `adrc.c` has `const float z3DecayRate = adrcRuntime->liftoff ? c->decayRate : c->gatedDecayRate;`. That distinction is exactly what matters for ADRC-026 — after a false liftoff detection the craft is still on the ground but the observer has already switched to the slow airborne `adrc_sigma_decay`. Worth also not implying the leak prevents windup: it bounds accumulation, but `-beta3 * errorEso` keeps driving z3 regardless.

```suggestion
- **Gated z3 decay** (`adrc_gated_z3_decay`, default 200 = 20/s) — z3 remains a leaky integrator in either latch state. While the liftoff latch is closed, its decay rate is `max(adrc_gated_z3_decay × 0.1, adrc_sigma_decay × 0.1, 1.0/s)`. At the default `200` this is 20/s (τ ≈ 0.05 s), so the 1.0/s floor does not bind; with `adrc_gated_z3_decay = 0` and the default `adrc_sigma_decay = 3`, it is 1.0/s (τ ≈ 1 s). Once the latch opens — even if it false-opens while the craft remains physically grounded — the observer switches to `adrc_sigma_decay`. The leak limits z3 accumulation but cannot guarantee that none occurs, because `-beta3 * errorEso` continues to drive z3.
```

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,174 @@
...
+| `adrc_liftoff_throttle`            | 40                 | 1–100             | Throttle % that alone confirms liftoff. See cross-reference note above — not derived from `adrc_hover_throttle`                                                                                              |
+| `adrc_liftoff_gyro_dps`            | 20                 | 1–255             | Sustained rotation (°/s, any axis) that alone confirms liftoff (toss-launch path)                                                                                                                            |
+| `adrc_liftoff_hold_ms`             | 25                 | 0–5000            | How long the rotation above must sustain before it counts                                                                                                                                                    |
+| `adrc_gated_z3_decay`              | 200 (=20.0)        | 0–2000            | z3 decay rate ×0.1 while the gate is closed (grounded) — never slower than `adrc_sigma_decay` (floored up to match if set lower), and never slower than an absolute 1.0/s floor regardless of either setting |
```
danusha2345: Same two fixes as the bullet above, condensed for the table: "while the gate is closed (grounded)" conflates latch state with ground contact, and the two floors are easier to read as the single three-way max the code actually computes.

```suggestion
| `adrc_gated_z3_decay`              | 200 (=20.0)        | 0–2000            | z3 decay rate ×0.1 while the liftoff latch is closed; effective decay is `max(adrc_gated_z3_decay × 0.1, adrc_sigma_decay × 0.1, 1.0/s)` |
```

docs/wiki/guides/current/ADRC.md
```diff
@@ -0,0 +1,174 @@
...
+- **`dshot_bidir = ON`** has been observed to cause hard gyro freezes (thousands of identical gyro samples, momentarily uncontrollable) on at least one board/firmware-base combination. If your craft flies erratically or "acts possessed," try `set dshot_bidir = OFF` (you lose the RPM filter; the dynamic notch still works).
+- **If the motors are audibly oscillating at idle right after arming, disarm immediately.** A high enough `adrc_wo` can make the craft self-oscillate on the ground at idle; the liftoff gate's gyro-only path can misread that oscillation as a real liftoff and open **while the craft is still grounded**, driving the motors toward saturation. Arm props-off first when trying a higher `wo`, and watch/listen closely as you raise it. See [Known limitations](#known-limitations) for the current status of this failure mode.
```
danusha2345, line 155: This bullet covers the self-oscillation entry path. Indoor logs from @8ksal8 posted to #15400 on 2026-07-29/30 (b5 `543f1a5ff`, wc 125 / wo 160 / b0 4000-3500-10000, hover 29, no wind) add a second one that a tester would not infer from it: **arming with ANGLE and airmode both active**, where the leveling loop demands rate off the craft's resting tilt at 0 % stick throttle.

On the comparison, note it is one arm per condition: ADRC with permanent airmode reached the 2047 rail in 60.9 % of the first 0.25 s while CLASSIC with permanent airmode had none and settled 0.46 s in — but the CLASSIC *box*-airmode arm also bounced (8.3 % of samples railed) and BOXAIRMODE state is not recorded in blackbox, so the ADRC-specific share is not quantified. The suggestion keeps the existing bullet unchanged and adds one that stays inside what the data supports.

```suggestion
- **If the motors are audibly oscillating at idle right after arming, disarm immediately.** A high enough `adrc_wo` can make the craft self-oscillate on the ground at idle; the liftoff gate's gyro-only path can misread that oscillation as a real liftoff and open **while the craft is still grounded**, driving the motors toward saturation. Arm props-off first when trying a higher `wo`, and watch/listen closely as you raise it. See [Known limitations](#known-limitations) for the current status of this failure mode.
- **Prefer arming an ADRC profile in ACRO, especially when using airmode.** If the craft is resting tilted, ANGLE commands a non-zero rate immediately. With `pid_at_min_throttle = ON`, the motors can answer at idle and make the craft skid or bounce even though stick throttle remains at 0%. If the liftoff latch is already open or false-opens during that motion, ADRC uses the configured airborne z3 decay while ground contact constrains the plant, so the observer can accumulate a disturbance estimate and compound the bounce. In one same-craft, one-arm-per-condition comparison, ADRC with permanent airmode reached motor saturation while CLASSIC with permanent airmode settled; however, the CLASSIC box-airmode arm also bounced and BOXAIRMODE state is not logged, so this comparison does not quantify the ADRC-specific share. Arm on level ground in ACRO and disarm immediately if the craft moves or oscillates at idle.
```

Original `.bbl`s, hashes, scripts and the full write-up: https://github.com/danusha2345/ADRC-betaflight/tree/master/docs/flight-test-analysis/pr15400-8ksal8-armbounce

danusha2345 (commented):
@bvandevliet Re-read the page at `73d2dd8` against firmware head `f4ce7dc135`. Five of the six earlier points are closed on the merits, and I agree with the follow-up commit: decoupling the page from fast-moving fork state is the right call, and nothing practical is lost — the concrete "which tag do I flash, and what will it reset" answer lives in the tracker's [How to help test](https://github.com/danusha2345/ADRC-betaflight/blob/master/docs/ADRC_REMEDIATION_TRACKER.md#how-to-help-test), which this page links twice, and I'll keep that current instead of the page. Thanks also for the `adrc.h` fix in `f4ce7dc135`.

What I re-verified against the head branch: every default and range in the CLI table matches (`adrc.c` defaults 60 / 100 (yaw 80) / 2000, `gyro_lpf 150`, `hover 35`, `sigma_decay 3`, `liftoff 40`/`20`/`25`, `gated_z3_decay 200`, `b0_scale_max 3`; ranges from `settings.c`), the armed/disarmed `pid_type` paragraph matches `pid_init.c` exactly (`pidResetIterm()` unconditional, `adrcResetGate()` only under `if (!ARMING_FLAG(ARMED))`), and the internal anchors resolve.

The sixth point — gated z3 decay — is still not right, and the way it is now phrased inherits a mistake I made in the original review comment, so the correction is on me. Two inline suggestions for that, plus one for the ADRC-026 caution, which currently describes only one of the two observed entry paths.

I've also retitled ADRC-026 in the tracker so its scope matches what the section already documents: **"Ground-constrained excitation can false-open the liftoff gate and drive zero-throttle z3 windup/runaway"** — the high-`wo` self-oscillation is one entry path; bench wind rocking and arming in ANGLE with airmode are two more.

danusha2345:
@bvandevliet Thanks — the applied text reads correctly, and the airmode/ANGLE bullet kept the caveat it needed. Three follow-ups; the first two are CodeRabbit's, and both are right, including one against my own wording.

**Line 43 — `z3` isn't leaky in both latch states.** "z3 remains a leaky integrator in either latch state" is wrong when `adrc_sigma_decay = 0`: `z3DecayRate = liftoff ? decayRate : gatedDecayRate`, and `decayRate` is `adrc_sigma_decay × 0.1`, so at the documented `0` setting the airborne path has no leak at all — the source comment says exactly that. Suggested:

> - **Gated z3 decay** (`adrc_gated_z3_decay`, default 200 = 20/s) — z3 remains an integrator in either latch state. While the liftoff latch is closed, its state decays at `max(adrc_gated_z3_decay × 0.1, adrc_sigma_decay × 0.1, 1.0/s)`; while airborne it uses `adrc_sigma_decay × 0.1`, which may be zero and therefore leave a pure integrator. …

(rest of the bullet unchanged). The table row needs nothing further: `adrc_gated_z3_decay` is additionally clamped to `ADRC_GATED_Z3_DECAY_MAX = 2000`, which equals the CLI maximum, so the documented formula holds across the whole settable range.

**Line 66 — the Simplified Tuning claim needs scoping, and that one's mine too.** `simplified_tuning.c` rewrites `gyro_lpf1_dyn_min_hz`, `gyro_lpf1_dyn_max_hz` and `gyro_lpf1_static_hz` from `simplified_gyro_filter_multiplier`, and ADRC consumes `gyro.gyroADCf` — so the gyro slider *does* change the ESO input, and "only ever recalculates `pid[axis].P/I/D/F` and `d_max[axis]`" is true only of the PID sliders. The D-term distinction I wrote still holds. I'd take the bot's wording essentially as-is.

**Line 157 — the efficiency bullet points at a number that isn't logged.** `energyCumulative` is not a flight-controller field: the S-frame carries only `flightModeFlags, stateFlags, failsafePhase, rxSignalReceived, rxFlightChannelsValid`. The column is synthesised by `blackbox_decode`, and it can be wrong by a wide margin — on a 33.3 s tester log it reports 115 mAh where the integral of the `amperageLatest` column in the same CSV is 51.18 mAh. I've reconstructed the decoder's arithmetic (three compounding defects: a `-1` seed on `lastFrameTime`, an already-scaled centiamp field pushed through the raw-ADC conversion, and an `int16_t` milliamp parameter that wraps the high-current samples) and I'm raising that against blackbox-tools separately — details and a repro script in [`pr15400-8ksal8-fall/`](https://github.com/danusha2345/ADRC-betaflight/tree/master/docs/flight-test-analysis/pr15400-8ksal8-fall). For the page, the safe form is to point at the raw samples instead:

> - If you want to compare efficiency (current draw) against classic PID, fly matched manoeuvres on both `pid_type`s and compare by independently integrating the logged `amperageLatest` samples. The `energyCumulative` column produced by `blackbox_decode` is not logged by the flight controller and can be wrong on recent Betaflight logs. The props-off heat finding above doesn't generalize to real flight either way.

PR #725, opened 2026-07-22
Add build server retention policy announcement

blckmn:
Adds a blog announcement covering the new build server retention policy: RC builds kept for 30 days, point releases kept for a maximum of 2 years, with the latest point release of every major version from 4.3 onwards always retained. Also notes that builds may be pulled sooner for community safety reasons, and that any version can be built locally from the tagged source.

### DIFF
--- a/blog/2026/07/22/Build Server Retention Policy.md
+++ b/blog/2026/07/22/Build Server Retention Policy.md
@@ -1,0 +1,26 @@
+---
+title: Build Server Retention Policy
+date: 2026-07-22
+authors: ['blckmn']
+---
+
+To keep our build server sustainable, we're introducing retention limits on the firmware builds we host.
+
+<!--truncate-->
+
+## What's Changing
+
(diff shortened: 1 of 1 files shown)


haslinghuis (approved):

nerdCopter (approved):
- approving -- is 2 years enough considering potentially releasing annually? (maybe 3?) we currently see some people preferring 4.5 over 2025 due issues.

osirisinferi:
> * is 2 years enough considering potentially releasing annually? (maybe 3?) we currently see some people preferring 4.5 over 2025 due issues.

How I interpret the text this 2 years is only for intermediate point releases. In theory the latest point release, which apparently is still kept indefinitely, should be issue-free, at least have less issues than the point release before that.

haslinghuis:
The **latest point release of every major version from 4.3 onwards** will always remain available, regardless of age. Even once the 2-year window has passed, you'll still be able to download the final build of each major release (4.3.x, 4.4.x, 4.5.x, 2025.12.x, and so on).

PR #752, opened 2026-08-11
docs: add guide for controlling Betaflight from a companion computer over MSP

vdmaas98:
Adds a guide page to docs/wiki/guides/current: Controlling Betaflight from a Companion Computer over MSP.

There is currently no guide covering RC input over MSP from a companion computer (Raspberry Pi, Jetson). msp_override appears only in the auto-generated CLI dumps and Defines.md, and serialrx_provider=MSP is only mentioned in passing, yet this is a common setup for autonomous flight projects and the two configurations have very different safety properties.

The page covers: both configurations (serialrx_provider=MSP for the bench, msp_override for flight), the silent-failure rules (mode switches must not sit on MSP-driven channels, masks must match), MSP_SET_RAW_RC payload and send-rate guidance, a props-off verification checklist, and a comparison table.

Everything is first-hand verified on Betaflight 4.5.2 (SPEEDYBEEF405V5, ELRS/CRSF receiver, Jetson companion over the USB VCP), including the behavior that surprised us: with MSP OVERRIDE active, a dead companion process does not fall back to the receiver. Betaflight holds the last MSP values indefinitely until the pilot flips the override switch (readback was still frozen minutes after the sender died). The guide documents that hold behavior and the companion-side consequence: never stop sending on an internal failure, command neutral values instead.

Also verified in the same session: MSP has zero authority while the override mode is inactive, AUX channels never follow MSP regardless (mask 15), attempts to set the ARM channel over MSP have no effect, and pilot reclaim via the switch is instant.

## Summary by CodeRabbit

* **Documentation**
  * Added a guide for controlling Betaflight from a companion computer.
  * Covers bench testing, receiver-backed in-flight override, configuration, arming, channel mapping, and link-failure behavior.
  * Includes message formatting, channel-value clamping, safety recommendations, build requirements, verification steps, and an operational comparison table.
  * Clarified that message length is determined by the highest-numbered masked channel.
  * Updated receiver documentation to describe sole-MSP and receiver-backed override modes, support for up to 18 channels, and their different safety properties.


vdmaas98:
Addressed all six findings. One correction and one nuance worth spelling out: configuration 1 now uses feature RX_MSP, since 4.5 has no MSP value for serialrx_provider (confirmed against a 4.5.2 CLI). On the hold behavior: I kept the indefinite-hold statement but scoped it per firmware version. It is measured behavior on 4.5.2 (values still held minutes after the sender died), while the 300 ms freshness fallback only entered with RX_MSP_RC_FRAME_FRESH_MS in the current releases. The guide now documents both, plus the msp_override_failsafe default.

nerdCopter (dismissed):

docs/wiki/guides/current/MSP-Companion-Computer-Control.md
```diff
@@ -0,0 +1,81 @@
...
+
+### 2. MSP override: the one to fly
+
+A real receiver supplies every channel. Betaflight substitutes MSP values for only the channels named by a bitmask, and only while the MSP OVERRIDE flight mode is active (a switch the pilot holds).
+
+```text
+set serialrx_provider = CRSF        # or whatever your real receiver uses
+set msp_override_channels_mask = 15
+save
+```
```
haslinghuis, line 30: Configuration 1's block is complete and copy-pasteable, so this one reads as if it were too — but `serialrx_provider` on its own is not a working receiver setup.

````suggestion
```text
# assumes a serial receiver is already working (feature RX_SERIAL plus a UART assigned to Serial RX)
set serialrx_provider = CRSF        # or whatever your real receiver uses
set msp_override_channels_mask = 15
save
```
````

File: docs/wiki/guides/current/MSP-Companion-Computer-Control.md [hunk 2]
```diff
@@ -0,0 +1,81 @@
...
+
+`msp_override_failsafe` defaults to `OFF`, which means losing the RC receiver link still triggers normal failsafe even while the override is active. Leave it `OFF` unless you specifically intend to fly on MSP without an RC link, which forfeits the safety properties this section is about.
+
+`15` is binary `1111`, meaning channels 1 to 4 (roll/pitch/throttle/yaw). The companion computer flies the aircraft; every AUX channel stays with the receiver, so ARM and the kill switch remain physical. Then, in the Modes tab, put ARM and MSP OVERRIDE on two real (receiver-driven) switches.
+
+Two rules that fail silently if violated:
+
+- Neither ARM nor MSP OVERRIDE may sit on an MSP-driven channel. Otherwise the companion process could arm itself, or trap itself in override with no way for the pilot to reclaim the sticks.
+- The mask on the flight controller and the mask the companion software assumes must be equal. A mismatch does not produce an error; it silently changes who controls which channel.
```
haslinghuis, line 39: The MSP OVERRIDE half of the first rule does not actually fail silently — the firmware protects against it. `validateAndFixConfig()` in `src/main/config/config.c:552-557` clears the mask bit for whichever AUX channel MSP OVERRIDE is assigned to, and `:548-550` removes the mode entirely when the mask is zero. You cannot trap yourself in override.

What *is* silent is that your saved mask gets rewritten underneath you, which feeds straight into the second rule. ARM has no equivalent protection, so that is the one that deserves the warning.

````suggestion
Two rules, only one of which the firmware protects you from:

- **ARM must not sit on an MSP-driven channel.** Nothing in the firmware prevents this, and if you allow it the companion process can arm the aircraft itself. Keep ARM on a receiver channel that the mask excludes.
- **The mask on the flight controller and the mask your companion software assumes must be equal.** A mismatch produces no error; it silently changes who controls which channel. Note that Betaflight adjusts the mask itself when you save: the bit for whichever AUX channel MSP OVERRIDE is assigned to is cleared automatically (so you cannot trap yourself in override), and if that leaves the mask at zero the MSP OVERRIDE mode is removed. Read the mask back after saving rather than assuming the value you set.
````

File: docs/wiki/guides/current/MSP-Companion-Computer-Control.md [hunk 3]
```diff
@@ -0,0 +1,81 @@
...
+
+When the pilot flips MSP OVERRIDE off, control returns to the receiver's live channels instantly. That switch is the pilot-takeover mechanism for the MSP-controlled channels (the ARM switch separately remains the kill authority), and what happens when the companion process stops sending depends on the firmware version:
+
+- Betaflight 4.5.x: the last received MSP values are held indefinitely. Verified on 4.5.2: no timeout, values still held minutes after the sender died. The aircraft keeps flying the last command until the pilot flips the switch.
+- Current releases (since the MSP RC freshness check, `RX_MSP_RC_FRAME_FRESH_MS = 300`): overridden channels fall back to the receiver 300 ms after MSP frames stop.
```
haslinghuis, line 44: The version boundary leaves out 2025.12.x, which behaves like 4.5.x, not like current master.

The freshness check landed in commit `61e7750` (#15217, 2026-06-27). `git tag --contains 61e7750` gives `2026.6.0-rc1` onwards, and `git show 2025.12.5:src/main/rx/msp.c` has no `FRESH` symbol at all. As written, a reader on 2025.12 concludes they get the 300 ms fallback — they get indefinite hold.

````suggestion
- Betaflight 4.5.x and 2025.12.x: the last received MSP values are held indefinitely. Verified on 4.5.2: no timeout, values still held minutes after the sender died. The aircraft keeps flying the last command until the pilot flips the switch.
- Betaflight 2026.6 and later (MSP RC freshness check, `RX_MSP_RC_FRAME_FRESH_MS = 300`): overridden channels fall back to the receiver 300 ms after MSP frames stop.
````

File: docs/wiki/guides/current/MSP-Companion-Computer-Control.md [hunk 4]
```diff
@@ -0,0 +1,81 @@
...
+
+Either way the pilot briefing is the same: on anything unexpected, flip MSP OVERRIDE off first, and do not rely on the firmware to hand control back for you on 4.5.x.
```
haslinghuis, line 46:
````suggestion
Either way the pilot briefing is the same: on anything unexpected, flip MSP OVERRIDE off first, and do not rely on the firmware to hand control back for you on 4.5.x or 2025.12.x.
````

File: docs/wiki/guides/current/MSP-Companion-Computer-Control.md [hunk 5]
```diff
@@ -0,0 +1,81 @@
...
+
+:::note
+
+MSP override is gated behind a build define (`USE_RX_MSP_OVERRIDE`). If `msp_override_channels_mask` is missing from your CLI, your target's default build does not include it. Use the cloud build system with the feature selected, or build with the define added.
+
+:::
```
haslinghuis, line 52: This note sends readers chasing a build option that does not exist.

`USE_RX_MSP_OVERRIDE` is defined unconditionally at preprocessor depth 0 in `src/main/target/common_pre.h:419` (and `:359` on the 4.5.5 tag). The only guard is `common_post.h:465`, which undefines it if `USE_RX_MSP` is absent — and `USE_RX_MSP` is unconditional too. Nothing in the `config` repo undefines either one, and there is no cloud build feature toggle for it. `msp_override_channels_mask` is present in every standard build.

````suggestion
:::note

MSP override is compiled into every standard Betaflight build: `USE_RX_MSP_OVERRIDE` is defined unconditionally in `src/main/target/common_pre.h`, and there is no cloud build option to select for it. If `msp_override_channels_mask` is missing from your CLI, you are on a cut-down or custom build.

:::
````

File: docs/wiki/guides/current/MSP-Companion-Computer-Control.md [hunk 6]
```diff
@@ -0,0 +1,81 @@
...
+
+## Sending MSP_SET_RAW_RC
+
+The message payload is N uint16 little-endian channel values in microseconds (1000 to 2000, center 1500), in your configured channel order (AETR by default: roll, pitch, throttle, yaw, then AUX1 and up).
```
haslinghuis, line 56: Two details worth adding: the handler rejects payloads over 18 channels with `MSP_RESULT_ERROR` (`src/main/msp/msp.c:3039-3042`), and in override mode values are *clamped* to `rx_min_usec`/`rx_max_usec` (885/2115 by default, `common_defaults_post.h:97-98`) rather than rejected — so out-of-range input fails silently.

````suggestion
The message payload is N uint16 little-endian channel values in microseconds, in your configured channel order (AETR by default: roll, pitch, throttle, yaw, then AUX1 and up). Use 1000 to 2000 with 1500 as center. Up to 18 channels are accepted; a longer payload is rejected with an MSP error. Values outside `rx_min_usec` to `rx_max_usec` (885 to 2115 by default) are silently clamped in override mode rather than rejected.
````

File: docs/wiki/guides/current/MSP-Companion-Computer-Control.md [hunk 7]
```diff
@@ -0,0 +1,81 @@
...
+
+Practical points:
+
+- Send continuously at a steady rate (25 to 50 Hz works well). In configuration 1, frames must keep arriving or the source is considered lost and the aircraft failsafes. In configuration 2, stopped frames mean held values on 4.5.x or a 300 ms fallback on current releases, per the section above.
```
haslinghuis, line 60:
````suggestion
- Send continuously at a steady rate (25 to 50 Hz works well). In configuration 1, frames must keep arriving or the source is considered lost and the aircraft failsafes. In configuration 2, stopped frames mean held values on 4.5.x and 2025.12.x, or a 300 ms fallback on 2026.6 and later, per the section above.
````

File: docs/wiki/guides/current/MSP-Companion-Computer-Control.md [hunk 8]
```diff
@@ -0,0 +1,81 @@
...
+- In override mode, send meaningful values only on the overridden channels and keep the rest at center. The flight controller ignores non-masked channels from MSP, but your own software should not rely on values it does not control.
```
haslinghuis, line 61: This is the one behaviour the guide is missing, and on pre-2026.6 firmware it is more dangerous than the hold you already document.

`rxMspFrameReceive()` (`src/main/rx/msp.c:63-70`) resets every channel past `channelCount` to zero. On 4.5.x and 2025.12.x, `rxMspOverrideReadRawRc()` then constrains that zero to `rx_min_usec` and returns it the instant BOXMSPOVERRIDE goes active. So flipping the override switch **before the companion has ever sent a frame**, or with a frame shorter than the mask, drops the masked channels — throttle included — to minimum immediately. The code comment added by #15217 in master documents exactly this as its motivation.

The converse is worth stating too: on 2026.6+, `rxMspIsRcChannelFresh()` returns false when `chan >= lastRxMspRcFrameChannelCount`, so short frames safely leave those channels with the receiver.

````suggestion
- Always send at least as many channels as your mask covers, in every single frame. Betaflight resets every channel past the end of your payload to zero. On 4.5.x and 2025.12.x that zero is then clamped to `rx_min_usec` (885 by default) and used, so a short frame — or enabling MSP OVERRIDE before your process has sent its first frame — drops the masked channels, throttle included, to minimum. On 2026.6 and later a channel beyond the last frame's length counts as stale and stays with the receiver.
- Beyond that, send meaningful values only on the overridden channels and keep the rest at center. The flight controller ignores non-masked channels from MSP, but your own software should not rely on values it does not control.
````

File: docs/wiki/guides/current/MSP-Companion-Computer-Control.md [hunk 9]
```diff
@@ -0,0 +1,81 @@
...
+- Never let your software simply stop sending in override mode: on 4.5.x going silent freezes the last command indefinitely, and even on current releases the aircraft flies your stale command for 300 ms. On any internal failure (planner hung, camera stalled), send neutral attitude values with reduced throttle and keep sending them until the pilot takes over.
```
haslinghuis, line 62:
````suggestion
- Never let your software simply stop sending in override mode: on 4.5.x and 2025.12.x going silent freezes the last command indefinitely, and even on 2026.6 and later the aircraft flies your stale command for 300 ms. On any internal failure (planner hung, camera stalled), send neutral attitude values with reduced throttle and keep sending them until the pilot takes over.
````

File: docs/wiki/guides/current/MSP-Companion-Computer-Control.md [hunk 10]
```diff
@@ -0,0 +1,81 @@
...
+- Any MSP implementation works, for example YAMSPy (Python), or raw struct packing over a serial library.
+
+## Verifying the setup before flight (props off)
+
+1. `MSP_RC` readback: send a distinctive pattern via `MSP_SET_RAW_RC` and read `MSP_RC` back. In configuration 1 you should see your values verbatim. In configuration 2 you should see the receiver's values until the pilot enables MSP OVERRIDE, then your values on the masked channels only.
+2. Confirm what a dead sender does on your firmware: with MSP OVERRIDE active, kill the companion process and read `MSP_RC` back. On 4.5.x the flight controller keeps the last MSP values until the pilot flips MSP OVERRIDE off; on current releases the overridden channels should return to the receiver after 300 ms. Flipping MSP OVERRIDE off must return the channels to the receiver instantly in both cases. In configuration 1, a dead sender means failsafe.
```
haslinghuis, line 68:
````suggestion
2. Confirm what a dead sender does on your firmware: with MSP OVERRIDE active, kill the companion process and read `MSP_RC` back. On 4.5.x and 2025.12.x the flight controller keeps the last MSP values until the pilot flips MSP OVERRIDE off; on 2026.6 and later the overridden channels should return to the receiver after 300 ms. Flipping MSP OVERRIDE off must return the channels to the receiver instantly in both cases. In configuration 1, a dead sender means failsafe.
````

File: docs/wiki/guides/current/MSP-Companion-Computer-Control.md [hunk 11]
```diff
@@ -0,0 +1,81 @@
...
+3. Confirm ARM authority: attempt to change the ARM channel over MSP in configuration 2. It must have no effect.
```
haslinghuis, line 69: "It must have no effect" reads as a firmware guarantee, but it only holds because mask 15 excludes AUX. The setting accepts up to `(1 << 18) - 1` (`src/main/cli/settings.c:926`) and nothing stops ARM's channel being masked. (Same over-generalisation in the PR description: "AUX channels never follow MSP regardless".)

````suggestion
3. Confirm ARM authority: with a stick-only mask such as `15`, attempt to change the ARM channel over MSP in configuration 2. It must have no effect. This follows from the mask, not from a firmware guarantee — the mask can cover AUX channels too, so never include ARM's channel in it.
````

File: docs/wiki/guides/current/MSP-Companion-Computer-Control.md [hunk 12]
```diff
@@ -0,0 +1,81 @@
...
+
+A configuration mismatch here does not fail loudly. It fails by the aircraft not responding, or responding when it should not. Check the settings programmatically at startup if your companion software supports it, or use a preflight checklist.
+
+## Choosing between them
+
+| | `feature RX_MSP` | MSP override |
+| --- | --- | --- |
+| Receiver needed | no | yes |
+| Who arms | companion computer | pilot (physical switch) |
+| Companion-death behavior | failsafe | 4.5.x: holds last values until pilot flips the switch; current releases: receiver after 300 ms |
```
haslinghuis, line 79:
````suggestion
| Companion-death behavior | failsafe | 4.5.x and 2025.12.x: holds last values until pilot flips the switch; 2026.6 and later: receiver after 300 ms |
````

haslinghuis (changes requested):
Useful page — the gap is real, and the structure (two configurations, the silent-failure rules, the props-off checklist) is the right shape for this material. I checked the claims against the firmware at `master`, `4.5.5`, `2025.12.5` and `2026.6.1`. Most hold up; two are wrong in ways that matter for a safety page, and one behaviour is missing.

**Verified correct:** `feature RX_MSP` is the right spelling (`cli.c:256`), MSP genuinely is not a `serialrx_provider` value (`lookupTableSerialRX`), `msp_override_failsafe` does default to `OFF`, mask bit 0 = channel 1, and `MSP_SET_RAW_RC` is code 200.

**Two things to fix before merge:**

1. The build-define note is wrong — `USE_RX_MSP_OVERRIDE` is in every standard build, and there is no cloud build option for it.
2. The 4.5.x vs "current releases" split omits 2025.12.x, which behaves like 4.5.x. That is the version most readers will actually be running.

**One thing missing:** channels past the end of your MSP payload are zeroed, and on pre-2026.6 firmware that zero is clamped to `rx_min_usec` and flown. Enabling MSP OVERRIDE before the companion has sent its first frame drops throttle to minimum instantly. Details inline.

**Non-blocking, and not inline because it is a different file:** nothing links to this page. `docs/wiki/guides/current/Rx.md:188` has a two-sentence "MultiWii serial protocol (MSP)" section that is the natural entry point, and `Modes.md:49` plus `docs/wiki/app/auxiliary-tab.md:60` both list MSP OVERRIDE with a bare one-liner. Worth a link from at least `Rx.md`. Also, the guide names `MSP_SET_RAW_RC` and `MSP_RC` without linking the MSP Protocol Reference.

Separately, `Rx.md:190` says MSP RX has "Only 8 channel support", which contradicts this guide's 18 — pre-existing, but a link between the two pages will make the inconsistency visible, so it may be worth correcting in the same PR.

CI-wise everything is clean: the H1 passes `check_title_case` (I ran it with the repo's own `title-case` dependency), `check_file_names` skips `docs/wiki`, and both fences are tagged `text`.

vdmaas98:
Thanks for the thorough review, and for the firmware references. All of it is in d8dc9e6.

Both errors are fixed. The build-define note now says MSP override is compiled into every standard build with no cloud build option, and the version split covers 2025.12.x alongside 4.5.x instead of treating the freshness check as current behaviour. That second one would have been the more harmful of the two, since 2025.12 is what most readers are running.

The short-frame clamping was genuinely missing and I had not understood it. It also explains something I hit on my own aircraft without knowing why, so I documented it and added a props-off check for it, because it is easy to trigger by enabling the override switch a moment too early.

On the silent-failure rules I have narrowed the first to ARM only and documented that Betaflight rewrites the mask on save, with the advice to read it back.

I took the non-blocking suggestions as well: Rx.md now links this page from its MSP section, the payload section gives the 18 channel limit and the clamping range, and MSP_SET_RAW_RC and MSP_RC link the protocol reference. I also corrected the "Only 8 channel support" line in Rx.md to 18 in the same commit, so let me know if you would rather that went as a separate PR.

haslinghuis:
@coderabbitai review

haslinghuis (approved):

PR #753, opened 2026-08-12
Docs: sync manufacturer documentation with config repo layout, current MCU targets and sensors

haslinghuis:
Brings the manufacturer documentation back in line with the current `betaflight/config` repository and firmware master. Every claim here was checked against firmware master and the config repository rather than carried over from the existing text; where the existing docs were wrong, they are corrected rather than extended.

## Config repository layout

`betaflight/config` moved from a flat `configs/<BOARD_NAME>/` layout to manufacturer-grouped `configs/<MANUFACTURER_ID>/<BOARD_NAME>/` in [config#1164](https://github.com/betaflight/config/pull/1164). Target selection is unchanged — `mk/config.mk` globs `configs/*/<BOARD>/config.h`, a flat match still wins, and a duplicate board name now fails the build with `Ambiguous CONFIG`.

- `creating-configuration` gains a Directory Layout section (grouped path, board-name uniqueness, the reserved `CUST` / `FOSS` / `COMM` / `LEGA` ids), a required-versus-optional defines table, the `SUPPORTED TARGET` header block (`REFERENCE:` / `DATE:` / `VERSION:` / `GROUP:`, including the no-leading-asterisk parsing rule), and the optional `config.c` / `config.mk` mechanisms with `OCTOSPI_FLASH_CHIP`.
- `requirements-for-submission-of-targets` now says where the config file goes, that config PRs must come from a custom branch, and that the reserved IDs need no new manufacturers-list entry.
- Stale flat paths fixed in `Customized-Version`, `Missing` and `Git`; four dead links to the deprecated `unified-targets` repo now point at `betaflight/config`.

## BMI270 shared-footprint alignment

New section 4.2.1.1 documents `ENABLE_BMI270_ALIGN_AS_ICM` from [#15484](https://github.com/betaflight/betaflight/pull/15484): what it does, that it is opt-in and defaults to `0`, that it needs a numeric value because it is evaluated with `#if`, that it is build-wide (so it rotates every BMI270 in a dual-gyro build), and that CW90 versus CW270 must be bench-verified before a config ships it. Released in 2026.6.1. Also added to `Defines` under a new section for build options without a `USE_` prefix.

## MCU platforms and targets

The guidelines only knew about F4, F7, G4, H7 and AT32F435. New section 4.4 covers all eight platforms and their valid `FC_TARGET_MCU` values — STM32 F4/F7/G4/H7/H5/C5/N6, AT32, APM32, PICO (RP2350), X32, ESP32, SITL — with the practical caveats: targets excluded from CI ship no release binaries, H730/H750 have 128 KB internal flash, N657 has no internal user flash, H757 is dual-core, 512 KB is the practical floor.

Section 3.2.1.3 is widened from "G4, H7, and AT32F435" to all families with flexible DMA routing (adding H5/C5/N6 GPDMA and X32M7), APM32 F4 is grouped with STM32 F4 for the bitbang errata, RP2350 and ESP32 are called out as not timer-driven, and the stale `F7X2` / `G4XX` target names are replaced with `F722` / `G474`.

## Sensors

- Added `USE_ACCGYRO_ICM56686` and `USE_BARO_SPA06_003`, which were missing from the define lists.
- Native gyro sample rates for the parts the looptime table never covered (6.4 kHz for ICM-45605/45686/56686, 6664 Hz for LSM6DSO, 9 kHz for ICM-20649).
- The EXTI pin requirement — omitting `GYRO_<n>_EXTI_PIN` silently drops the board to polled SPI reads with no DMA.
- Previously undocumented defines: `USE_GYRO_CLKIN` / `GYRO_<n>_CLKIN_PIN` (ICM-42688-P only, fixed 32 kHz), the gyro SPI clock overrides, `ENABLE_42686_EXTENDED_RANGE`, and the baro/mag bus and address defines.
- A warning that `USE_BARO` and `USE_MAG` are **not** implied by chip defines, unlike the rangefinder defines — a config listing only a chip compiles no driver at all.
- `BARO_I2C_INSTANCE` / `MAG_I2C_INSTANCE` must always be declared: they fall back to `I2C_DEVICE`, which is `I2CINVALID` unless the config defines it.
- BMP580/581 default to I2C address 0x47; QMC5883L is 0x0D and QMC5883P 0x2C; the IST8310 auto-detects only at 0x0E, with `set mag_i2c_address = 12` as the field recovery path.
- Two known firmware issues are documented as caveats: `USE_ACCGYRO_IIM42652` is missing from the SPI detection table, and the SPI paths for BMP580/581 are guarded on a define nothing sets.

## Corrections to existing docs

- `creating-configuration` documented `USE_GYRO_1_ALIGN` and `ALIGN_1_GYRO_ROLL/PITCH/YAW`. None of those symbols exist — zero occurrences in the firmware and zero across the 626 configs. The section now uses `GYRO_1_ALIGN`, and `GYRO_1_ALIGN_ROLL` / `_PITCH` / `_YAW` or `GYRO_1_CUSTOM_ALIGN` (mutually exclusive), and gained the gyro wiring and multi-gyro rules.
- The same file's `FC_TARGET_MCU ATF435M` example was not a real target, and the `AIRBOTF435` board it referenced no longer exists in the config repository. Replaced with the current `configs/ATRY/ATSTARTF435` values.
- Gyro alignment is now defined as the IMU orientation relative to the board's forward arrow — including boards whose arrow sits at 45 degrees to the PCB edges — while `DEFAULT_ALIGN_BOARD_YAW` is described as a user-level mounting decision that does not normally belong in a board config.
- Removed the obsolete STM32F303CC datasheet entry from `Flight-Controller-Design` and pointed it at the new platform table.
- Rebuilt the `Supported Sensors` wiki page, which was badly out of date (it ended at "ICM40625" and "DSP310"), from the current firmware drivers.
- Fixed heading numbering in the guidelines (a duplicated 3.1.4.1 and a mis-levelled 3.1.5).

## Review follow-ups (out of scope here)

Found while verifying, not fixed in this PR because they live in other repositories:

- `betaflight/config` — `configs/TEBS/TBS_LUCID_RPI_FC/config.h` defines `USE_BARO_BMP390`, which does not exist in the firmware; the BMP388 driver already detects the BMP390 by chip ID, so it should be `USE_BARO_BMP388`.
- `betaflight/config` — `configs/MTKS/MATEKF411WTE/config.h` uses `USE_GYRO_SPI_BMI270` / `USE_ACC_SPI_BMI270` instead of `USE_ACCGYRO_BMI270`.
- `betaflight/config` — 41 configs set `DEFAULT_ALIGN_BOARD_YAW`; if the alignment guidance in this PR is the project's position, some of those are compensating in the wrong place.
- `betaflight` — `USE_ACCGYRO_IIM42652` is absent from the `icm426xxSpiDetect` guard in `accgyro_mpu.c`, and `barometer_bmp5xx.c` guards its SPI paths on `USE_BARO_SPI_BMP5XX`, which nothing defines.

`npm run check` and `npm run build` both pass with no warnings.


docs/development/manufacturer/creating-configuration.md
```diff
@@ -188,14 +278,18 @@ CW270_DEG_FLIP
...
-#define ALIGN_1_GYRO_PITCH DECIDEGREES
-#define ALIGN_1_GYRO_YAW DECIDEGREES
+#define GYRO_1_ALIGN_ROLL  0
+#define GYRO_1_ALIGN_PITCH 0
+#define GYRO_1_ALIGN_YAW   1350
 ```
 
+The per-axis defines and `GYRO_1_CUSTOM_ALIGN` are mutually exclusive — using both produces a build error.
+
+Where the whole board is rotated relative to the airframe, use `DEFAULT_ALIGN_BOARD_YAW` rather than baking the rotation into the gyro alignment. Magnetometer alignment follows the same pattern with `MAG_ALIGN` and `MAG_ALIGN_ROLL` / `MAG_ALIGN_PITCH` / `MAG_ALIGN_YAW`.
```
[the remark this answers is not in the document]
haslinghuis: Addressed, but not as proposed — `GYRO_1_CUSTOM_ALIGN` is a real symbol, so I kept it and documented the interaction instead.

In `src/main/pg/gyrodev.c` (master):

```c
#ifndef GYRO_1_ALIGN
#ifdef GYRO_1_CUSTOM_ALIGN
#define GYRO_1_ALIGN            ALIGN_CUSTOM
#else
#define GYRO_1_ALIGN            CW0_DEG
#endif
#endif // GYRO_1_ALIGN

#ifndef GYRO_1_CUSTOM_ALIGN
#define GYRO_1_CUSTOM_ALIGN     SENSOR_ALIGNMENT_FROM_STD(GYRO_1_ALIGN)
#else
STATIC_ASSERT(GYRO_1_ALIGN == ALIGN_CUSTOM, "GYRO_1_ALIGN and GYRO_1_CUSTOM_ALIGN mixed");
#endif
```

So the per-axis `GYRO_1_ALIGN_ROLL/PITCH/YAW` defines compose `GYRO_1_CUSTOM_ALIGN` themselves, which in turn makes `GYRO_1_ALIGN` default to `ALIGN_CUSTOM` — adding `#define GYRO_1_ALIGN ALIGN_CUSTOM` is permitted but not required. What genuinely fails is combining a preset with per-axis values: the `STATIC_ASSERT` then fires. That is the failure mode manufacturers actually hit, so the text now says the per-axis defines imply `ALIGN_CUSTOM`, that an explicit `GYRO_1_ALIGN` must be `ALIGN_CUSTOM`, and that `GYRO_1_CUSTOM_ALIGN` is the pre-composed alternative and mutually exclusive with the per-axis form.

docs/wiki/guides/current/Supported-Sensors.md
```diff
@@ -2,45 +2,103 @@
...
+| Nooploop TOFSense               | UART      | Rangefinder                    |
+| HC-SR04                         | GPIO      | Ultrasonic rangefinder, legacy |
+
+## Blackbox Flash
 
-## Range Finder
+| Chip family                             | Interface               | Notes                                                              |
+| :-------------------------------------- | :---------------------- | :----------------------------------------------------------------- |
+| Winbond W25Qxx and JEDEC-compatible NOR | SPI                     | Covered by the M25P16 driver's JEDEC ID table                      |
+| Winbond W25Q128FV                       | SPI / QuadSPI / OctoSPI |                                                                    |
```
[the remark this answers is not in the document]
haslinghuis: Partly applied. The QuadSPI half is right and is now fixed; the OctoSPI half is not — `flash_w25q128fv.c` does have an OctoSPI path.

From master, the driver's top-level guard:

```c
#if defined(USE_FLASH_W25Q128FV) && (defined(USE_QUADSPI) || defined(USE_OCTOSPI))
```

and it branches on `USE_OCTOSPI` throughout (`#if defined(USE_QUADSPI) ... #elif defined(USE_OCTOSPI)` at several call sites, plus an `#if defined(USE_OCTOSPI)` block further down). So OctoSPI stays on that row.

What I did change, since your QuadSPI point was correct: `flash_m25p16.c`, `flash_w25n.c` and `flash_mt29f.c` all carry `USE_QUADSPI` paths, so the generic NOR, W25N and MT29F rows now read SPI / QuadSPI rather than SPI. The W25Q128FV row is now "QuadSPI / OctoSPI" with a note that the same part on plain SPI is handled by the M25P16 driver, which is what the driver guards actually describe.

docs/development/manufacturer/creating-configuration.md
```diff
@@ -55,9 +89,25 @@ The license information is used to identify the license of the firmware. The con
...
+| `FC_TARGET_MCU`   | MCU family/type. The build derives `TARGET` from this and fails if it is missing or cannot be parsed.       |
+| `BOARD_NAME`      | Unique board name. Must match the config's leaf directory name (`configs/<MANUFACTURER_ID>/<BOARD_NAME>/`). |
+| `MANUFACTURER_ID` | Four-letter manufacturer code. Must match the parent directory the config lives under.                      |
+
+Commonly used optional defines:
+
+| Define              | Purpose                                                                                                                                                         |
+| :------------------ | :-------------------------------------------------------------------------------------------------------------------------------------------------------------- |
+| `SYSTEM_HSE_MHZ`    | External oscillator frequency in MHz; sets `HSE_VALUE` for the clock tree. Required on boards with an HSE crystal, omit if running off the internal oscillator. |
+| `FC_VMA_ADDRESS`    | Sets the load/run (VMA) address. Its presence switches the build to an EXST (bootloader-relocated) image. Rare.                                                 |
```
osirisinferi: Isn't it a bit strange to have a "rare" define listed in a list of "Commonly used defines"? Recommend leaving it out.

docs/development/manufacturer/creating-configuration.md
```diff
@@ -55,9 +89,25 @@ The license information is used to identify the license of the firmware. The con
 #define MANUFACTURER_ID AIRB
 ```
 
+These three defines are mandatory — both the firmware build and the cloud build API derive the target's identity from them:
+
+| Define            | Purpose                                                                                                     |
+| :---------------- | :---------------------------------------------------------------------------------------------------------- |
+| `FC_TARGET_MCU`   | MCU family/type. The build derives `TARGET` from this and fails if it is missing or cannot be parsed.       |
+| `BOARD_NAME`      | Unique board name. Must match the config's leaf directory name (`configs/<MANUFACTURER_ID>/<BOARD_NAME>/`). |
+| `MANUFACTURER_ID` | Four-letter manufacturer code. Must match the parent directory the config lives under.                      |
```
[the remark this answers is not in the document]
haslinghuis: Applied in `0e6ecdb`. Confirmed the finding first: `ATF435M` appears nowhere in the firmware tree, the AT32 targets are `AT32F435G` / `AT32F435M`, and the `AIRBOTF435` board in the example no longer exists in the config repository at all.

The example now uses a real, current config (`configs/ATRY/ATSTARTF435/config.h`):

```c
#define FC_TARGET_MCU AT32F435M
#define BOARD_NAME ATSTARTF435
#define MANUFACTURER_ID ATRY
```

I also added a line stating that `FC_TARGET_MCU` must match one of the firmware's target names, linking to the new supported-platform table, since an unmatched value fails the build in `mk/config.mk`.

docs/development/manufacturer/requirements-for-submission-of-targets.md
```diff
@@ -4,10 +4,12 @@ The following new requirements for pull requests adding new targets or modifying
...
+2. New manufacturers are required to be added to the [manufacturers list](https://github.com/betaflight/config/blob/master/Manufacturers.md). Your four-letter manufacturer ID is also the name of the directory your configs live in, so it must be registered before the first target is submitted.
 
-3. For any new targets, a definition file needs to be submitted to [betaflight/config](https://github.com/betaflight/config). See the [instructions](https://betaflight.com/docs/development/manufacturer/creating-configuration) for how to create a configuration file. Ensure you adhere to the [Config Target Guidance](config-target-guidance) also.
+3. For any new targets, a definition file needs to be submitted to [betaflight/config](https://github.com/betaflight/config), placed at `configs/<MANUFACTURER_ID>/<BOARD_NAME>/config.h`. Board names must be unique across the whole repository. See the [instructions](https://betaflight.com/docs/development/manufacturer/creating-configuration) for how to create a configuration file. Ensure you adhere to the [Config Target Guidance](config-target-guidance) also.
```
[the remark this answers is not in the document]
haslinghuis: Applied in `0e6ecdb`, with one nuance: the reserved IDs are not an exception to the manufacturers list — `CUST`, `FOSS`, `COMM` and `LEGA` are already entries in `Manufacturers.md`. So the contradiction is about needing a *new* entry, not about being absent from the list.

Item 2 now reads that submissions with no registered vendor use one of those reserved IDs and need no new entry, which keeps the registration requirement intact for actual vendors.

docs/wiki/guides/current/Supported-Sensors.md
```diff
@@ -2,45 +2,103 @@
...
+
+:::note
+
+Having a driver does not make a part a good choice. Recommendations for new hardware designs are in the [Manufacturer Design Guidelines](/docs/development/manufacturer/manufacturer-design-guidelines), and parts that are no longer accepted for new designs are called out there.
+
+:::
+
+## Gyroscope and Accelerometer
+
+Gyroscopes must be connected over SPI. I2C gyros are not supported for new designs.
```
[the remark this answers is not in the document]
haslinghuis: Applied in `0e6ecdb` — the line now reads "For new designs, gyroscopes must be connected over SPI. I2C gyros are only supported on legacy boards", which no longer reads as contradicting the MPU-6050 row.

docs/wiki/guides/current/Supported-Sensors.md
```diff
@@ -2,45 +2,103 @@
...
+| QMC5883L | I2C       | Normal axis orientation, works well; default address 0x0D                   |
+| QMC5883P | I2C       | Default address 0x2C                                                        |
+| HMC5883L | I2C / SPI |                                                                             |
+| LIS2MDL  | I2C       |                                                                             |
+| LIS3MDL  | I2C       |                                                                             |
+| MMC560X  | I2C       |                                                                             |
+| AK8963   | I2C / SPI | Also reachable through the MPU925X pass-through                             |
+| AK8975   | I2C       | Legacy                                                                      |
+| IST8310  | I2C       | Not recommended — non-standard axis orientation requires a custom alignment |
+
```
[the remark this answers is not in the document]
haslinghuis: Applied in `0e6ecdb`. Confirmed against `compass_ist8310.c` (`IST8310_MAG_I2C_ADDRESS 0x0E`) and the existing guidance in the design guidelines, which already covered the CAD-selected alternatives.

The row now states that auto-detection uses 0x0E and that the other addresses (0x0C, 0x0D, 0x0F) require `mag_i2c_address` to be set manually, alongside the existing custom-alignment warning.

haslinghuis:
All six review comments have been answered inline and the review is now stale — `0e6ecdb` addresses every one of them.

| Finding | Outcome |
| --- | --- |
| `FC_TARGET_MCU ATF435M` invalid | Fixed — example replaced with the real `configs/ATRY/ATSTARTF435` values, plus a note that the value must match a firmware target |
| Custom gyro alignment selector | Fixed differently — `GYRO_1_CUSTOM_ALIGN` is a real symbol; documented that the per-axis defines imply `ALIGN_CUSTOM` and that mixing a preset with per-axis values trips the `STATIC_ASSERT` |
| Reserved manufacturer IDs | Fixed — noted that `CUST`/`FOSS`/`COMM`/`LEGA` are already in `Manufacturers.md` and need no new entry |
| Absolute SPI gyro statement | Fixed — scoped to new designs |
| IST8310 I2C address behaviour | Fixed — 0x0E auto-detect and the `mag_i2c_address` override added to the row |
| Flash interface claims | Partly fixed — QuadSPI added to the M25P16, W25N and MT29F rows; OctoSPI kept on W25Q128FV, whose driver is guarded on `USE_QUADSPI \|\| USE_OCTOSPI` |

@coderabbitai review

docs/development/manufacturer/creating-configuration.md
```diff
@@ -188,14 +280,18 @@ CW270_DEG_FLIP
...
-#define ALIGN_1_GYRO_PITCH DECIDEGREES
-#define ALIGN_1_GYRO_YAW DECIDEGREES
+#define GYRO_1_ALIGN_ROLL  0
+#define GYRO_1_ALIGN_PITCH 0
+#define GYRO_1_ALIGN_YAW   1350
 ```
 
+The per-axis defines set the alignment to `ALIGN_CUSTOM` on their own, so `GYRO_1_ALIGN` can be left out. If it is given, it must be `ALIGN_CUSTOM` — combining a preset such as `CW180_DEG` with per-axis values fails the build. `GYRO_1_CUSTOM_ALIGN` is the alternative, pre-composed form of the same setting, and is mutually exclusive with the per-axis defines.
+
+Where the whole board is rotated relative to the airframe, use `DEFAULT_ALIGN_BOARD_YAW` rather than baking the rotation into the gyro alignment. Magnetometer alignment follows the same pattern with `MAG_ALIGN` and `MAG_ALIGN_ROLL` / `MAG_ALIGN_PITCH` / `MAG_ALIGN_YAW`.
```
osirisinferi: IMO here we should make clear that the gyro alignment is meant for the physical IMU chip alignment relative to the boards front (arrow?) and that board alignment should be used for I dunno, FCs purely used in BNF drones with a non-standard FC orientation?

Sometimes you'd see AIOs with +45 or -45° board alignments, even if the AIO has an arrow pointing to the front, where said front is at a 45° offset from the board square sides. But IMO this is incorrect: with the board fitted in such a way as that the arrow is pointing forward on the quad, the board alignment should be 0°. Not + or -45°. This should be fixed in the gyro alignment.

IMO board alignment is most likely purely a user setting if the user decides to build the FC differently into a quad than where the arrow points to. Not an FC specific hardware setting that should be present in a `config.h`.

docs/development/manufacturer/manufacturer-design-guidelines.md
```diff
@@ -542,6 +553,8 @@ Additionally, there are no RC ecosystems that are actively developing a supporte
 
 Note that the use of gyros such as the BMI270 lowers the gyro loop rate from 8kHz to 3.2kHz and therefore may be advantageous for F411 designs.
 
+Targets on the newer platforms (STM32 H5, C5, N6, RP2350, X32M7 and APM32) are not listed above. Treat H5, C5, N6 and X32M7 designs as H7-class for these purposes, and APM32 F4 as F405-class, but agree the defaults with the Betaflight team as part of the target review — see [4.4](#44-supported-mcu-platforms-and-targets) for the status of each platform.
```
osirisinferi: > Treat H5, C5, N6 and X32M7 designs as H7-class for these purposes (&hellip;)

I'm not so sure about H5 and C5. The H5 has about the same CoreMark rating as an F7 and the C5 a bit more than half of that.

docs/development/manufacturer/manufacturer-design-guidelines.md
```diff
@@ -592,6 +612,71 @@ Define at least one gyro and one accelerometer. Gyroscopes must communicate via
...
+
+This flag is off by default so that no existing board changes its frame of reference. Only set it when you have confirmed the resulting orientation on hardware — changing it for an existing board name will silently rotate every user's craft. CW90 versus CW270 is layout dependent, so bench-verify pitch, roll and yaw signs with the BMI270 fitted before shipping a config that sets it. On boards where a BMI270 and a second gyro are fitted and used together, note that the flag rotates every BMI270 in the build — confirm the two gyros still agree.
+
+:::
+
+Available from Betaflight 2026.6.1 onwards ([PR #15484](https://github.com/betaflight/betaflight/pull/15484)).
+
+Other BMI270 constraints that matter when sharing a footprint with the ICM-42688-P:
+
+- Maximum SPI clock is 10 MHz, well below what the ICM parts tolerate, so the layout must be good enough for the slower part.
```
osirisinferi: > so the layout must be good enough for the slower part

What do you mean by this? 🙂 PCB layout? B/c usually PCB layouts become more important with *faster* clock speeds. With slower clock speeds, the PCB routing is usually less strict.

docs/development/manufacturer/manufacturer-design-guidelines.md
```diff
@@ -641,6 +736,16 @@ Define a barometer only if physical present on the board. Betaflight strongly re
 #define USE_BARO_SPI_BMP581
 ```
 
+The bus and, where needed, the address are declared separately — `BARO_I2C_INSTANCE` (defaults to `I2C_DEVICE`, so a baro on any other bus must say so), `BARO_SPI_INSTANCE` and `BARO_CS_PIN` for SPI parts, and `DEFAULT_BARO_I2C_ADDRESS` (decimal) where the fitted part cannot use its default address. `DEFAULT_BARO_DEVICE` pins which driver is used, and the BMP085 additionally needs `BARO_XCLR_PIN`.
```
osirisinferi: > defaults to `I2C_DEVICE`, so a baro on any other bus must say so

`I2C_DEVICE` actually defaults to `I2CINVALID`, so it's not just baro's on any other bus: its instance should always be defined, as there is no valid default. Unless I'm reading the code incorrectly that is 🙂

osirisinferi (commented):
Nice large update! 👍 A few small notes though, maybe not all that relevant, but hopefully useful 🙂

docs/development/manufacturer/creating-configuration.md
haslinghuis: Agreed — a row whose own description says "Rare" does not belong under "Commonly used". Removed in `b5cb6ed`; the table now lists only `SYSTEM_HSE_MHZ` and `TIMER_PIN_MAPPING`.

docs/development/manufacturer/creating-configuration.md
haslinghuis: You're right, and my wording recommended the opposite of what it should have. Fixed in `b5cb6ed`.

The section now says the gyro alignment describes how the IMU chip is oriented relative to the board's own forward marking (the silkscreened arrow), and calls out the 45-degree case explicitly: that offset belongs in the gyro alignment, so a board installed with its arrow facing the front of the craft needs no further correction from the user.

`DEFAULT_ALIGN_BOARD_YAW` is now described separately as the board being mounted at an angle to the airframe — a user decision rather than a property of the hardware — and therefore not something that normally belongs in a board's `config.h`.

Worth noting for a possible follow-up: 41 configs currently set `DEFAULT_ALIGN_BOARD_YAW`. I have not touched those, but if the guidance above is the project's position then some of them are probably compensating in the wrong place.

docs/development/manufacturer/manufacturer-design-guidelines.md
haslinghuis: Fair — that was too coarse. Fixed in `b5cb6ed`.

The line now reads that the N6 and X32M7 are H7-class in compute, the H5 is closer to an F7 despite being a newer part, the C5 is lower again at roughly half the H5, and APM32 F4 tracks F405. That matches the ordering you describe: H563 is an M33 at 250 MHz against an M7 at 216 MHz for the F722, and the C5 runs at 144 MHz.

docs/development/manufacturer/manufacturer-design-guidelines.md
haslinghuis: You're right, that sentence was muddled and, as written, backwards. Fixed in `b5cb6ed`.

What I meant was a bus-throughput point, not a PCB routing one: the BMI270 caps SPI at 10 MHz where the ICM parts run at 24 MHz, so on a board that populates either part the bus runs at whatever the fitted chip allows and the BMI270 variant gets less headroom for gyro reads. It now says that instead.

docs/development/manufacturer/manufacturer-design-guidelines.md
haslinghuis: You're reading it correctly, and my wording was misleading. Fixed in `b5cb6ed`, in this file and in both the baro and mag sections of the design guidelines.

`src/main/drivers/bus_i2c.h`:

```c
#ifndef I2C_DEVICE
#define I2C_DEVICE I2CINVALID
#endif
```

and `common_defaults_post.h` falls `BARO_I2C_INSTANCE` / `MAG_I2C_INSTANCE` back to `I2C_DEVICE`. Only a handful of configs define `I2C_DEVICE` to a real bus (the SPRACINGH7 boards and a couple of Nucleos), so for everything else there is no usable default. The config repo agrees in practice: 427 configs declare `BARO_I2C_INSTANCE` against 416 that enable `USE_BARO`.

The text now says the instance must always be given, rather than only for a sensor on a non-default bus.

docs/development/manufacturer/manufacturer-design-guidelines.md
```diff
@@ -662,7 +767,9 @@ Define a magnetometer only if physical present on the board. Betaflight strongly
...
+For the QMC5883, prefer the chip-specific `USE_MAG_QMC5883L` or `USE_MAG_QMC5883P` for the part actually fitted. `USE_MAG_QMC5883` is a hybrid define retained for backwards compatibility — the firmware expands it to enable both drivers (see `common_post.h`) — so existing configs that use it continue to build. Note that the two parts use different default I2C addresses: 0x0D for the QMC5883L and 0x2C for the QMC5883P.
+
+As with the barometer, `USE_MAG` itself must be defined for any compass driver to be built. The bus is declared with `MAG_I2C_INSTANCE`, which like the barometer has no usable default and must always be given, or `MAG_SPI_INSTANCE` and `MAG_CS_PIN` for an SPI part, and `MAG_I2C_ADDRESS` (decimal) overrides the address where the default cannot be used. `MAG_INT_EXTI` declares a data-ready interrupt pin, which is only used when `USE_MAG_DATA_READY_SIGNAL` is defined. Board-relative orientation is set with `MAG_ALIGN`, or per axis with `MAG_ALIGN_ROLL` / `MAG_ALIGN_PITCH` / `MAG_ALIGN_YAW`.
```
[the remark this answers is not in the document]
haslinghuis: Applied in `0f7e76d` — the define/setting distinction was worth making explicit.

Confirmed in master: `src/main/sensors/compass.c:121-151` only uses the uppercase `MAG_I2C_ADDRESS` build define to seed `compassConfig->mag_i2c_address` (default `0`, i.e. auto-detect), and `src/main/cli/settings.c:826` exposes the lowercase `mag_i2c_address` as a runtime `HARDWARE_VALUE` setting.

§4.2.4 now spells out that the build define only sets the shipped default, that a user can change the runtime setting from the CLI, and gives the recovery path for an IST8310 at 0x0C (`set mag_i2c_address = 12` then `save`), cross-linked to §3.1.3.1 which already covered the address rules and the 12/13/15 values.

One thing I did not assert: whether the setting is reachable from the Configurator GUI. I could not verify that from this checkout, so the text says only that it is changeable from the CLI.

haslinghuis:
@coderabbitai please open issues on the relevant repo's for review follow-ups (out of scope here)

Found while verifying, not fixed in this PR because they live in other repositories:

- betaflight/config — configs/TEBS/TBS_LUCID_RPI_FC/config.h defines USE_BARO_BMP390, which does not exist in the firmware; the BMP388 driver already detects the BMP390 by chip ID, so it should be USE_BARO_BMP388.
- betaflight/config — configs/MTKS/MATEKF411WTE/config.h uses USE_GYRO_SPI_BMI270 / USE_ACC_SPI_BMI270 instead of USE_ACCGYRO_BMI270.
- betaflight/config — 41 configs set DEFAULT_ALIGN_BOARD_YAW; if the alignment guidance in this PR is the project's position, some of those are compensating in the wrong place. (SKIP THIS ONE as it needs tested on hardware and will be fixed if required)
- betaflight — USE_ACCGYRO_IIM42652 is absent from the icm426xxSpiDetect guard in accgyro_mpu.c, and barometer_bmp5xx.c guards its SPI paths on USE_BARO_SPI_BMP5XX, which nothing defines.

osirisinferi (approved):

nerdCopter (approved):

